PR #1, opened 2018-08-16
Add service profile RFC

adleong:

### DIFF
--- a/service_profiles.md
+++ b/service_profiles.md
@@ -1,0 +1,111 @@
+# Service Profiles
+
+The RFC introduces the notion of service profiles.  A service profile is an
+object that describes characteristics that a service may have.  Throughout
+this RFC we will extensively use the REST Service Profile as an example.  The
+REST Service Profile describes the characteristics of an HTTP service that
+obeys the tenants of [REST]().
+
+Service profiles are useful because this allows Linkerd to behave differently
+depending on the characteristics of the service to which it is making requests.
+For example, if Linkerd sends an HTTP GET request to a service that has the
+REST profile, the request is guaranteed to be idempotent and therefore is safe
+for Linkerd to retry.  The types of characteristics that could potentially be
+described by a service profile includes:
+
+* success classification
+* retryability/idempotence
+* retry budgets
+* rate limits
+* ACLs
+* latency objectives/timeouts
+* success rate objectives
+* etc.
+
+Each service can be associated with at most one profile.  While we may one day
+wish to explore extending or overriding profiles, this is currently out of
+scope.  It is important to note that even though it is a service that is
+associated with a profile, this can influence the behavior of callers of that
+service.  In many cases the service and its caller may exist in different
+namespaces and/or have different owners.
+
+Service profiles are designed to be built in an incremental and extensible way.
+The initial implementation will contain very few characteristics but will be
+expanded over time.
+
+## Service Profile Definition
+
+While some characteristics such as retry budgets apply to the service as a
+whole, many characteristics vary depending on the request.  For example,
+idempotence may vary depending on the request's HTTP method or timeouts might
+vary depending on the request's HTTP path.  
+
+To accommodate this, a service profile contains some top level characteristics 
+that apply to the whole service and an ordered list of request policy objects
+which define the characteristics which apply to certain classes of requests.
+
+```
+message ServiceProfile {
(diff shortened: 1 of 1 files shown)


service_profiles.md
```diff
@@ -0,0 +1,111 @@
...
+      // idempotent requests
+      condition: Any(Method(GET), Method(PUT), Method(DELETE), Method(HEAD), Method(TRACE))
+      success_condition: Status < 500
+      retry_condition: Always
+    },
+    RequestPolicy {
+      // fallback, (non-idempotent)
+      condition: Always
+      success_condition: Status < 500
+      retry_condition: Never
```
seanmonstar, line 90: I think I can see why having a `Never` makes sense as something *higher* priority, (say, PUTs with a 500 are dangerous in my system, but all other idempotent methods with any 5xx are safe), but is it necessary to have a fallback? I'd assume that if the list of policies has been exhausted and nothing matched, the proxy can just assume that it cannot retry.

adleong: Yeah, depending on what the default service profile is, this fallback might be redundant.  What is listed there in the fallback is probably a good default (for HTTP).

seanmonstar: I think I forgot that this was also supposed to be used to give better metrics, so even though the fallback would not be to retry, it could still configure how to classify a success.

adleong: Yeah, though the success function defined there is probably the default success function for HTTP anyway.

service_profiles.md
```diff
@@ -0,0 +1,111 @@
...
+which responses should be considered successful and another response match
+condition that describes which responses should be considered retryable if they
+are not successful.
+
+```
+message RequestPolicy {
+  RequestMatch condition = 1;
+  ResponseMatch success_condition = 2;
+  ResponseMatch retry_condition = 3;
+  Duration timeout = 4;
```
seanmonstar, line 70: How do multiple policies with different timeouts get applied? Is the timeout supposed to be the max amount of time before the proxy should just return a 504?

service_profiles.md
```diff
@@ -0,0 +1,111 @@
...
+vary depending on the request's HTTP path.
+
+To accommodate this, a service profile contains some top level characteristics
+that apply to the whole service and an ordered list of request policy objects
+which define the characteristics which apply to certain classes of requests.
+
+```
+message ServiceProfile {
+  repeated RequestPolicy request_policies = 1;
+  RetryBudget retry_budget = 2;
```
seanmonstar, line 50: Is `RetryBudget` where backoff strategy would be included? Should that be specced out now, or are we punting on any sort of backoff?

adleong: Good question.  I had planned on punting on backoff for now.  But yes, this is probably where it would go.

service_profiles.md
adleong: First matching request policy wins.

I've been thinking more about this timeout field and I'm a bit worried that we might be conflating a true timeout (after which we should return a 504) and a retry timeout (after which we should attempt no more retries).

Setting a default true timeout is dangerous because it can potentially break applications that have very long/slow responses (e.g. streaming responses).  A retry timeout is much safer because in the worst case we just don't retry.

What do you think of changing this field name to `retry_timeout`?

seanmonstar: I dont know if this is different in finagle, but in the proxy, the existing max timeout doesn't trigger for streaming responses, it's only for waiting for the headers. Once the headers have been received, and the body pipe is setup, no more timeout is applied at all.

service_profiles.md
```diff
@@ -0,0 +1,111 @@
+# Service Profiles
+
+The RFC introduces the notion of service profiles.  A service profile is an
+object that describes characteristics that a service may have.  Throughout
+this RFC we will extensively use the REST Service Profile as an example.  The
```
wmorgan, line 5: Maybe this is overly pedantic, but aren't the idempotency semantics around GETs vs POSTs vs whatever a function of the HTTP spec, not REST? If so, would it be more accurate to call this the "Conforming HTTP service profile" or something like that?

seanmonstar: Yes, method idempotency is part of RFC 7230 (HTTP/1.1) at least, probably HTTP/1.0 also?

admc:
Feel free to put this in the new format and submit if needed!

PR #7, opened 2020-02-18
Automated Releases

kleimkuhler:
[Rendered](https://github.com/kleimkuhler/rfc/blob/kleimkuhler/automated-releases/design/0001-automated-releases.md)

PR #8, opened 2020-02-18
Remove RFC steps

kleimkuhler:
## Problem Statement

Before opening #7 and reading through the RFC instructions, the `Step 1` and
`Step 2` separation of RFC proposals were not very clear.

It is explained that RFCs should complete `Step 1` in one pull request and
then complete `Step 2` in a follow-up pull request. This was a little
confusing because it is also explained that once an RFC is accepted and
active, `.. RFCs should not be substantially changed. Only very minor changes
should be submitted as amendments`.

Additionally, after I opened #7 it seemed like I should include the `Design
Proposal` from the start because we already have pursed feedback on the
process.

## Solution

This proposes removing the `Step 1` and `Step 2` RFC separation so that once
an RFC is merged, follow-up pull requests are not needed in order to add
further details.

It is explained that one should consider not filling in the details until they
have pursued feedback on `Problem Statement`, but once that is done it then it
should be completed in separate commits of the *same pull request*.

Signed-off-by: Kevin Leimkuhler <kevin@kleimkuhler.com>


grampelberg:
@kleimkuhler the reason we've got those separate is so that the person who's filling out the problem statement doesn't have to be the one who does the design details. Imagine someone like me posting an RFC that everyone's behind. I don't want to go and add design details to the RFC itself.

Maybe we should state that it is possible to have the problem statement merged without design details? It doesn't guarantee anything other than this being a problem we'd like to work on in the future?

kleimkuhler:
> the person who's filling out the problem statement doesn't have to be the one who does the design details. Imagine someone like me posting an RFC that everyone's behind. I don't want to go and add design details to the RFC itself

@grampelberg I think an RFC design around this would lead to RFC's being merged because the community thinks it's a good idea, but then it goes stale after that.

If someone takes the time to open up a PR that describes the RFC, I feel like they should also be familiar enough with Linkerd to have an idea about the design details as well? I think most discussion is going to be solely *about* the implementation, so if that discussion does not happen before an RFC merges is it possible that it may not be worth it after all?

kleimkuhler:
It just felt a little odd in #7 to discuss the `Problem Statement`, but then following the directions not mention anything about the design details and have to do that in a separate pull request.

grampelberg:
Hmm, that's a good point. I'm trying to find a middle ground between a fast issue on the repo and something that someone *really* wants to see implemented.

kleimkuhler:
@grampelberg Maybe opening an issue in linkerd2 that is a `Feature Request` should still be the place for someone who doesn't want to invest the time into thinking about the details.

This RFC process is when that issue has enough discussion for someone to actually say "here is the feature/change, and this is how it will be implemented".

grampelberg:
I actually want it to be in the RFC repo, but we could have two separate docs in the repo. One for problem statements and one for implementation details. Would that work?

kleimkuhler:
Which reinforces the ability in an RFC to link to the issue (as directed to in the instructions) for further background on why this may be wanted in the first place.

grampelberg:
I'd like for the problem statement to be in *a* repo because:

- Commenting on issues is a huge PITA.
- The bar for working on a major feature must be higher than just writing up an issue.
- There needs to be discussion and details sorted out in the problem statement which are easy to miss in issue templates.

kleimkuhler:
> Commenting on issues is a huge PITA.

With discussion happening on a PR that is nice to be able to have discussions on specific wording and conversations that are separate from other conversations. That's a good point.

I think PR discussions can still become quite linear in terms of quoting, but I see the value in being able to open an RFC without having to also specify the details.

So what would changing the RFC process look like after this change then? A pull request is opened that adds an `.md` to an `rfc/accepted` directory?

Then once that occurs, a pull request is opened that moves the `.md` to an `rfc/active` directory? It additionally adds details to `Design Proposal`?

grampelberg:
Could just have a `problem` and `implementation` directory which each piece goes into. They definitely don't need to be identical docs, just makes it easier to comment on and refer to.

admc:
I'd like to keep it one doc so that once it's active people don't have to know to look in two separate places to follow the story. But I think the two directory idea moving the RFC from accepted to active would totally work for me.

kleimkuhler:
I also would like to keep it on doc since referring to two separate locations could be a little annoying. But the rest of this sounds good.

I will update this PR to reflect that process and maybe can then update #7 to work through that instead.

admc:
👍

kleimkuhler:
@admc @grampelberg Okay this has been updated to the discussion above. Let me know your thoughts on the structure changes.

Some drawbacks I see with this is there is a a lot of "process" moving markdown files around with what is in the "problem" step and "design" step. The directories are essentially acting as labels for what step an RFC is in.

I think this is still valuable as it may be helpful to see what accepted RFCs are in particular steps, but maybe all this really needs is more explicit instructions about `Step 1` and `Step 2`.

The benefits of this are that the `active` directory is not a combination of RFCs that have either been accepted and not designed, or already through the design phase.

admc (approved):
👍

PR #10, opened 2020-03-09
Thrift RFC

pnovotnak:
At Reddit we use Thrift as our internal RPC protocol. We'd love to use Linkerd2, but we need it to be able to support Thrift.

This RFC is for adding the same level of Thrift support we see with HTTP – routing, retries and metrics.

### DIFF
--- a/problem/0000-rfc-template.md
+++ b/problem/0000-rfc-template.md
@@ -1,0 +1,71 @@
+- Contribution Name: Thrift Support
+- Implementation Owner: Peter Novotank <peter.novotnak@reddit.com>, Linkerd2 Authors
+- Start Date: 6/3/2020
+- Target Date: 6/15/2020
+- RFC PR: [linkerd/rfc#9](https://github.com/linkerd/rfc/pull/9)
+- Linkerd Issue: [linkerd/linkerd2#0000](https://github.com/linkerd/linkerd2/issues/0000)
+- Reviewers: (Who should review the code deliverable? ex.@olix0r)
+
+# Summary
+
+[summary]: #summary
+
+Linkerd2-proxy and control plane support for the Thrift protocol.
+
+# Problem Statement (Step 1)
+
+[problem-statement]: #problem-statement
+
+- Add protocol support to the linkerd2-proxy for at least TFramedTransport and TBinaryProtocol 
+- Support routing, retries and metrics for Thrift.
+
+In many organizations, Thrift is used as the internal RPC protocol. In order to support 
+this use-case, linkerd2 needs to support routing and rate limiting. These features will 
+allow us to support advanced routing, as well as allow users to configure their mesh to 
+prevent retry storms and provide metrics.
+
+# Design proposal (Step 2)
+
+[design-proposal]: #design-proposal
+
+**Note**: This should be completed as part of `Step 2`.
+
+This is the technical portion of the RFC. Explain the design in sufficient detail that:
+
+- Its interaction with other features is clear
+- It is reasonably clear how the contribution would be implemented
+- Corner cases are dissected by example
+- Dependencies on libraries, tools, projects or work that isn't yet complete
+- Use Cases
+- Goals
+- Non-Goals
+- Deliverables
+
+# Prior art
+
+[prior-art]: #prior-art
+
+Discuss prior art, both the good and the bad, in relation to this proposal.
+A few examples of what this can include are:
+
+- Does this functionality exist in other software and what experience has their community had?
+- For other teams: What lessons can we learn from what other communities have done here?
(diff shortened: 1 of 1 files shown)


problem/0000-rfc-template.md
```diff
@@ -0,0 +1,71 @@
...
+
+[summary]: #summary
+
+Linkerd2-proxy and control plane support for the Thrift protocol.
+
+# Problem Statement (Step 1)
+
+[problem-statement]: #problem-statement
+
+- Add protocol support to the linkerd2-proxy for at least TFramedTransport and TBinaryProtocol
```
adleong, line 19: Linkerd can already transparently proxy any TCP traffic, so we should be more specific about what protocol support for thrift actually means.  What are the concrete list of features that Linkerd should support for thrift traffic?

problem/0000-rfc-template.md
```diff
@@ -0,0 +1,71 @@
...
+[summary]: #summary
+
+Linkerd2-proxy and control plane support for the Thrift protocol.
+
+# Problem Statement (Step 1)
+
+[problem-statement]: #problem-statement
+
+- Add protocol support to the linkerd2-proxy for at least TFramedTransport and TBinaryProtocol 
+- Support routing, retries and metrics for Thrift.
```
adleong, line 20: How should routing work for thrift messages? For HTTP message, Linkerd will look at the host or authority header and if it is formatted like a kubernetes service (x.y.svc.z), it will route to that service.  Thrift messages don't encode any routing information in the message itself; how should we decide where to send them?

adleong, line 20: Which metrics, specifically, should Linkerd produce for thrift traffic?

adleong, line 20: For HTTP, retryability is configured through service profiles.  How should service profiles interact with thrift traffic? A service profile defines a list of routes which are typically matched based on HTTP method and path regexes.  This doesn't translate very well to thrift where there are no HTTP methods or paths.  Should we extend service profiles in some way to accommodate thrift traffic?  Perhaps by allowing routes to match on the thrift service name or thrift method name? Or should we use some different mechanism to signal thrift retryability?

adleong, line 20: For HTTP, Linkerd currently can only retry requests which do not contain a request body. While we plan to support retrying requests with bodies eventually, this will require substantial changes to the way that buffering is done in the Linkerd proxy.

This is relevant because all thrift requests have bodies.  If retries are a requirement for thrift support then the aforementioned buffering changes become a prerequisite.

adleong, line 20: Response classification is fairly easy in HTTP since we assume (by default) that any response with status code 5XX is a failure and any other response code is a success.  Thrift does not have status codes like this so it is less clear how to determine if a response should be considered successful or not.  How do you go about calculating the success rate of a thrift service?

problem/0000-rfc-template.md
adleong, line 19: This RFC doesn't mention mTLS.  Should that be a requirement also?

problem/0000-rfc-template.md
```diff
@@ -0,0 +1,71 @@
...
+
+In many organizations, Thrift is used as the internal RPC protocol. In order to support 
+this use-case, linkerd2 needs to support routing and rate limiting. These features will
```
adleong, line 23: Linkerd does not currently support rate-limiting.

problem/0000-rfc-template.md
```diff
@@ -0,0 +1,71 @@
...
+# Problem Statement (Step 1)
+
+[problem-statement]: #problem-statement
+
+- Add protocol support to the linkerd2-proxy for at least TFramedTransport and TBinaryProtocol 
+- Support routing, retries and metrics for Thrift.
+
+In many organizations, Thrift is used as the internal RPC protocol. In order to support 
+this use-case, linkerd2 needs to support routing and rate limiting. These features will 
+allow us to support advanced routing, as well as allow users to configure their mesh to
```
adleong, line 24: By advanced routing, do you mean SMI traffic splitting, or something else?

adleong (commented):
Thanks for kicking this off, @pnovotnak! Thrift support in Linkerd would be really exciting. The next steps here are to use this RFC document to drill down into some of the technical details and define more specifically what the requirements are for thrift support.

I'll ask some questions on this PR that will hopefully get us going in the right direction. Definitely feel free to hop into Linkerd slack if you need any help or want to discuss.

cpretzer:
Hi @pnovotnak just checking in on this to see if you've had a chance to review the comments that @adleong made.

This would be a really useful feature and we'd love to work with you to get it done.

admc:
We are interested in continuing the conversation, so if you find time, please re-open!

PR #11, opened 2020-03-16
CI health metrics

alpeb:
- Contribution Name: ci-health-metrics
- Implementation Owner: @alpeb, @kleimkuhler
- Start Date: 2020-03-16
- Target Date: 2020-04-03
- RFC PR: [linkerd/rfc#11](https://github.com/linkerd/rfc/pull/11)
- Linkerd Issue:
  [linkerd/linkerd2#4176](https://github.com/linkerd/linkerd2/issues/4176)
- Reviewers: @grampelberg

# Summary

[summary]: #summary

This RFC proposes a service that would monitor the results from each CI run. It
should allow us to catch consistently flaky tests and have an overview of the CI
in terms of success rates and time performance.

# Problem Statement (Step 1)

CI jobs inevitably fail occasionally, either because of real problems with the
code, or because of test flakiness. We'd like to stay informed about what tests
are being persistently flaky, without having to dig into the CI run results
themselves. This will allow us to properly prioritize work to fix those
failures.

Besides identifying the most common failures, by surfacing error metrics from the
CI jobs we'll also have at our disposal an overview of CI's performance, in
terms of things like success rates and time taken by the jobs. This information
can be aggregated at the global, workflow and job levels.

The first problem to address will be to expose the CI jobs results, which
currently are only available through Github's UI. With access to that data in
the appropriate format, we'll then be able to determine how to persist and
present it to operators.

# Design proposal (Step 2)

Will be detailed once Step 1 gets approved.

# Prior art

[prior-art]: #prior-art

Github's Workflows and Checks API have only until recently become beta, so it's
expected there aren't well known tools that leverage that information.

I did find the [BuildPulse](https://github.com/marketplace/buildpulse/) Github
App, that specializes on tracking flaky tests in Github Actions. I couldn't make
it work in a Linkerd2 fork. And presumably it will also require the changes
described in Phase 1 above for it to properly give us actionable data instead of
data about the generic error messages we currently have. BuildPulse is closed
source and a paid service.

# Unresolved questions

[unresolved-questions]: #unresolved-questions

Hosting of the Prometheus exporter and Grafana dashboard that this solution
would require. Presumably AKS?

# Future possibilities

[future-possibilities]: #future-possibilities

Once we expose the CI metrics, we open the door to present this data however we
wish in the future.

### DIFF
--- a/problem/0000-ci-health-metrics.md
+++ b/problem/0000-ci-health-metrics.md
@@ -1,0 +1,81 @@
+- Contribution Name: ci-health-metrics
+- Implementation Owner: @alpeb, @kleimkuhler
+- Start Date: 2020-03-16
+- Target Date: 2020-04-03
+- RFC PR: [linkerd/rfc#11](https://github.com/linkerd/rfc/pull/11)
+- Linkerd Issue:
+  [linkerd/linkerd2#4176](https://github.com/linkerd/linkerd2/issues/4176)
+- Reviewers: @grampelberg
+
+# Summary
+
+[summary]: #summary
+
+This RFC proposes a service that would monitor the results from each CI run. It
+should allow us to catch consistently flaky tests and have an overview of the CI
+in terms of success rates and time performance.
+
+# Problem Statement (Step 1)
+
+[problem-statement]: #problem-statement
+
+CI jobs inevitably fail occasionally, either because of real problems with the
+code, or because of tests flakiness. We'd like to be able to be informed about
+what tests are being persistently flaky, without having to dig into the CI run
+results themselves. This will allow us to properly prioritize work to fix those
+failures.
+
+Besides identifying the most common failures, by surfacing error metrics from the
+CI jobs we'll also have at our disposal an overview of CI's performance, in
+terms of things like success rates and time taken by the jobs. This information
+can be aggregated at the global, workflow and jobs levels.
+
+## Phase 1
+
+CI errors are surfaced as annotations in Github's Workflows and Checks APIs.
+Currently our unit and integration tests (both in go and javascript) are not
+exposing errors as annotations. Therefore when a job fails, only a generic error
+is exposed and not the real error message. One first phase would require changes
+in the test suite for it to output errors in a format suitable for Github in
+order to expose them as annotations and through the Workflows and Checks APIs.
+
+CI leverages scripts from `/bin` that will also need to be touched for them to
+properly format their failure output as well (probably through some wrapper
+function).
+
+## Phase 2
+
+Once the CI metrics are surfaced through Github's API, additional functionality
+should be built in order to consume and present that data to operators.
+
+# Design proposal (Step 2)
+
+Will be detailed once Step 1 gets approved.
+
+# Prior art
+
+[prior-art]: #prior-art
+
+Github's Workflows and Checks API have only until recently become beta, so it's
+expected there aren't well known tools that leverage that information.
+
+I did find the [BuildPulse](https://github.com/marketplace/buildpulse/) Github
+App, that specializes on tracking flaky tests in Github Actions. I couldn't make
+it work in a Linkerd2 fork. And presumably it will also require the changes
+described in Phase 1 above for it to properly give us actionable data instead of
+data about the generic error messages we currently have. BuildPulse is closed
+source and a paid service.
+
+# Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+Hosting of the Prometheus exporter and Grafana dashboard that this solution
+would require. Presumably AKS?
+
+# Future possibilities
+
+[future-possibilities]: #future-possibilities
+
+Once we expose the CI metrics, we open the door to present this data however we
+wish in the future.
--- a/problem/0000-ci-health-metrics.md
+++ b/problem/0000-ci-health-metrics.md
@@ -25,10 +25,8 @@ are being persistently flaky, without having to dig into the CI run results
 themselves. This will allow us to properly prioritize work to fix those
 failures.
 
-Besides identifying the most common failures, by surfacing error metrics from the
-CI jobs we'll also have at our disposal an overview of CI's performance, in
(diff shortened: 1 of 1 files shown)


problem/0000-ci-health-metrics.md
```diff
@@ -0,0 +1,81 @@
...
+what tests are being persistently flaky, without having to dig into the CI run
+results themselves. This will allow us to properly prioritize work to fix those
+failures.
+
+Besides identifying the most common failures, by surfacing error metrics from the
+CI jobs we'll also have at our disposal an overview of CI's performance, in
+terms of things like success rates and time taken by the jobs. This information
+can be aggregated at the global, workflow and jobs levels.
+
+## Phase 1
```
grampelberg: I might be jumping to implementation details too soon, but have you considered junit output instead of something github specific?

grampelberg (approved):
I'm into it! Tracking run times and reporting on that is pretty interesting, though I suspect more of a intellectual interest than anything else. Even without a dashboard to visualize everything, getting granular reporting saved historically is going to help out a ton.

problem/0000-ci-health-metrics.md
```diff
@@ -0,0 +1,81 @@
...
+This RFC proposes a service that would monitor the results from each CI run. It
+should allow us to catch consistently flaky tests and have an overview of the CI
+in terms of success rates and time performance.
+
+# Problem Statement (Step 1)
+
+[problem-statement]: #problem-statement
+
+CI jobs inevitably fail occasionally, either because of real problems with the
+code, or because of tests flakiness. We'd like to be able to be informed about
```
kleimkuhler: nit: `test flakiness`

kleimkuhler: nit: `We'd like to stay informed about ..`?

problem/0000-ci-health-metrics.md
kleimkuhler: I'm also curious about this. I know with GH workflows we have already tied CI to GH specific functionality, but is there a more general output we can use so that we try to remain somewhat independent from where our integration tests are running?

kleimkuhler (approved):
This looks good to me as well. Left a few nits and echo'd Thomas's question!

problem/0000-ci-health-metrics.md
```diff
@@ -0,0 +1,81 @@
...
+failures.
+
+Besides identifying the most common failures, by surfacing error metrics from the
+CI jobs we'll also have at our disposal an overview of CI's performance, in
+terms of things like success rates and time taken by the jobs. This information
+can be aggregated at the global, workflow and jobs levels.
+
+## Phase 1
+
+CI errors are surfaced as annotations in Github's Workflows and Checks APIs.
```
olix0r: I hope I'm not being too nitpicky here but this all sounds like solution-space stuff to me, rather than problem-statement. If we're splitting the problem into two chunks, what are the problems that are solved in the first step? Parsing output of scripts is an implementation detail of solving what problem?

problem/0000-ci-health-metrics.md
alpeb: Unfortunately, currently GH Actions don't allow exporting test result data, like other tools do in xUnit format or what have you. Besides their own UI the only way to surface that data is through:

1) Annotations, that are then accessible through the Workflows and Checks API. This requires that our tests output annotation strings, which are formatted like `::error file=app.js,line=10,col=15::MessageHere`
2) Have our tests output any format (like xUnit reports), that we can store as a GH artifact.

My initial plan (not outlined in this Step 1 of the RFC), was to do (1) above and then have a Prometheus export consume that and then have Grafana show it. I showed a sample end result during the demo last Friday:
image
(Note the data backing that dashboard is all fake, I haven't hooked any of this yet)

As for the xUnit route, we could pull that from the artifacts, but I haven't been able to find any compelling tool that could leverage that, specifically for showing historical trends (or persistently flaky tests), which I believe is what we're after. I did find [UnitTH](http://junitth.sourceforge.net/) but that looks pretty dated, and there's also [allure2](https://github.com/allure-framework/allure2), which seems would require more effort to integrate. Do you have something else in mind?

problem/0000-ci-health-metrics.md
alpeb: Phase 1 would solve the problem that test results aren't currently being exposed for consumption anywhere besides the Github UI.
Once data is exposed, Phase 2 would solve the problem of how to consume and present that data.
Does that make sense, or am I missing something else?

olix0r:
Yep. I think this section should enumerate the problems, and the design proposal below should discuss workflows, checks, etc.

Oliver Gould < ver@buoyant.io >

problem/0000-ci-health-metrics.md
olix0r: Yep. I think this section should enumerate the problems, and the design proposal below should discuss workflows, checks, etc.

grampelberg (approved):
I'm excited to see the design!!

olix0r:
+1 on moving this into design.

problem/0000-ci-health-metrics.md
```diff
@@ -0,0 +1,68 @@
...
+CI jobs inevitably fail occasionally, either because of real problems with the
+code, or because of test flakiness. We'd like to stay informed about what tests
+are being persistently flaky, without having to dig into the CI run results
+themselves. This will allow us to properly prioritize work to fix those
+failures.
+
+Besides identifying the most common failures, by surfacing error metrics from the
+CI jobs we'll also have at our disposal an overview of CI's performance, in
+terms of things like success rates and time taken by the jobs. This information
+can be aggregated at the global, workflow and job levels.
```
olix0r: This is a somewhat handwavey description of the metrics that we wish to collect. Can we get more specific? I.e., independent of implementation, what do we want to measure?

alpeb: Ok, I've clarified in my last push what is that we want to measure.

admc (approved):
👍

PR #12, opened 2020-03-20
Add isolated integration tests

kleimkuhler:
[Rendered](https://github.com/kleimkuhler/rfc/blob/kleimkuhler/isolated-integration-tests/problem/0002-isolated-integration-tests.md)


problem/0002-isolated-integration-tests.md
```diff
@@ -0,0 +1,102 @@
...
+- Corner cases are dissected by example
+- Dependencies on libraries, tools, projects or work that isn't yet complete
+- Use Cases
+- Goals
+- Non-Goals
+- Deliverables
+
+# Prior art
+
+[prior-art]: #prior-art
```
alpeb, line 95: I think the intention of this section is to refer to third-party solutions we might look into to use, leverage or get inspired on  to solve the problems stated in this RFC. E.g. does Prow use this approach to parallel testing?

Maybe the deficiencies you listed here should be extracted into the problem-statement section above.

alpeb (commented):
Thanks @kleimkuhler :+1:

We talked about how test isolation might allow parallel execution, but I'm thinking now that might increase a bit too much the complexity. These are some things that come to mind that would need to get addressed:
- How will results be collected?
- Is that for both CI and local?
- How many clusters to run in parallel?
- If we allow for the current sequential approach to still be available, how do we switch between modes?

One problem that I didn't see stated explicitly is the ability to launch some given test on demand, which isolation would help solving.

Would it make sense to leave parallelism out of scope for the moment, or given your exploratory work so far is that something worth tackling from the get-go?

kleimkuhler:
> We talked about how test isolation might allow parallel execution, but I'm thinking now that might increase a bit too much the complexity.

@alpeb Thanks for the insight. I agree that parallel execution should be something that we consider out of scope for the initial RFC. It was something I was thinking about when writing this up, but I did ultimately at it to the `Future Possibilities` section because of similar concerns you listed.

I can change the wording so that the RFC makes less of an emphasis on parallel test execution.

> One problem that I didn't see stated explicitly is the ability to launch some given test on demand, which isolation would help solving.

Yes I was very clear in this that launching a specific test on demand would be easier than it current is. I briefly mentioned it in `Prior Art` as a bad property of the existing tests, but I should mention it in `Problem Statement` as something that change.

kleimkuhler:
@alpeb I removed the focus on parallel execution, as well as talked less about solution and more about the problem.

I provided Kubernetes using KinD as an example of prior art, and moved some of the points I previously made in that section up the `Problem Statement`.

problem/0002-isolated-integration-tests.md
```diff
@@ -0,0 +1,97 @@
...
+
+# Summary
+
+[summary]: #summary
+
+It will be easier to isolate integration tests by more than just namespace.
+With stronger isolation, running the tests will more reliable since they will
+not be as dependent on a complete cleanup of previous test runs. Additionally,
+running individual integration tests will more accessible; currently this
+requires knowledge about how the internal scripts work.
```
olix0r: This summary assumes that the reader understands details of the current integration test system. "more reliable" in relation to what? What does it mean to "be easier to isolate integration tests by more than just namespace?" Why do we care about that? What does it change for the project?

What do you think about something more along the lines of:

```suggestion
[summary]: #summary
Linkerd has an extensive suite of integration tests, but these tests have become fragile as they've
grown. This fragility manifests as false-positive test failures, which impact development velocity, and
it also inhibits the implementation of multi-cluster and cluster-level failure tests.

This proposal outlines a plan to address these problems by isolating Linkerd's integration tests to
not rely on external state/configuration.
```

problem/0002-isolated-integration-tests.md
olix0r: What are other projects using?

problem/0002-isolated-integration-tests.md
```diff
@@ -0,0 +1,97 @@
...
+- Use Cases
+- Goals
+- Non-Goals
+- Deliverables
+
+# Prior art
+
+[prior-art]: #prior-art
+
+### KinD
```
olix0r, line 97: What about K3s/k3d?

olix0r (commented):
I have more feedback but some quick notes:

problem/0002-isolated-integration-tests.md
kleimkuhler: Answering your question about "Why do we care about that?", the problem this RFC is trying to solve is that integration tests are not currently isolated enough. Isolation is not a problem made worse by the growth of the integration tests, it's always been there.

I think poor code and "hacks" to mock certain components are examples of things that come out of expanding tests and trying to quickly get something working, but test isolation has always been an issue.

For example, false-positives are more a symptom of the flakiness we see with tests that rely on a regex for log lines, and those log lines may change with the version of Kubernetes that we are testing with.

I can reword the summary to try and communicate that better, but I think your example touches more on the flakiness that we sometimes see, and I was not under the the impression that test flakiness is a goal of this RFC.

Is that helpful at all in explaining why I wrote the problem statement that way?

problem/0002-isolated-integration-tests.md
```diff
@@ -0,0 +1,151 @@
...
+
+The integration test setup checks require that certain conditions are
+satisfied by the given cluster. A surprising condition is that no pre-existing
+Linkerd installation resource may exist; if it does then it is deleted. These
+conditions exist because the test suite is constrained to a single
+cluster--running each test in its own namespace.
+
+After this change, the test suite should isolate tests by cluster. This will
+allow for each test to require any pre-existing state that is required without
+affecting later runs or existing installations.
```
alpeb, line 60: Did you mean each test will setup any initial state it requires?

problem/0002-isolated-integration-tests.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Extensibility
+
+Extending integration tests for more complex installation setups is not
+accessible.
+
+After this change, the testing process should be extendable in order to test
+new features being added to Linkerd. This can include things like checking the
+behavior of multi-cluster installations as well as the handling of errors in
+cluster failure scenarios.
```
alpeb, line 73: Besides having tests trigger/setup their own clusters, do you have something else in mind to enable the kind of tests you mention here? Just curious.

problem/0002-isolated-integration-tests.md
```diff
@@ -0,0 +1,151 @@
...
+  with existing Linkerd installations
+- Tests that require complex cluster states would be runnable by default
+
+
+Should the cluster creation tool provide the ability for tests to setup their
+own clusters?
+
+- If the tool used for cluster creation provided a programmatic way for tests
+  to setup their own clusters instead of depending on scripts to set the
+  state, the tests could set finer grained cluster properties.
```
alpeb, line 136: :+1:

problem/0002-isolated-integration-tests.md
kleimkuhler: > Did you mean each test will setup any initial state it requires?

Yes and no. My wording here is not very specific about how the state is set because the problem phase does not propose whether tests should programmatically setup their state, or if that state is already existing from running scripts like what currently happens.

It looks like for the time being, tests will still depend on scripts setting up the state since there is not a great option for programmatically doing this, but I didn't want to rule that out in this statement.

problem/0002-isolated-integration-tests.md
kleimkuhler: > Besides having tests trigger/setup their own clusters

I don't think we have the right tools for that now. k3s is currently working on adding it, but it's probably a "future possibility" for this.

Here, I am talking more about how tests can be extended once there is less of a constraint on isolation. It may be better read as "once the integration tests are isolated by cluster, *then* we can write integration tests that have multi-cluster installations".

alpeb (approved):

problem/0002-isolated-integration-tests.md
```diff
@@ -0,0 +1,151 @@
...
+on in several scripts.
+
+Isolating through KinD clusters will provide stronger isolation than namespace
+because creating a new a cluster for each test ensures previous runs do not
+affect the current installation.
+
+It may not be possible to always rely on KinD if it is not installed, which is
+why namespace isolation should still be available.
+
+### k3d (and k3s)
```
olix0r, line 110: The other big benefit to k3d, imo, is that its new version offers an SDK so that tests could integrate directly without having to require the system has the correct version of k3d installed.

olix0r (commented):
I think what you have here sets the stage pretty well, but I don't have a good sense, yet, of what we're proposing, concretely. What are the next steps for this? Is it worthwhile try to port a single test to use cluster isolation strategy so we can get a sense of how much work is involved? It feels like we basically are trying to develop a library/framework for writing tests against Kubernetes... how do we focus this? If this one done ideally, what would an example test look like?

olix0r:
@kleimkuhler I may be overcomplicating some of this, due to my ignorance of the existing integration test infra. We should use the design section of this to get into the specifics of what actually needs to change. It should address at least: How do integration tests work today? Specifically: what are the components of the existing system? What pieces need to change?

kleimkuhler:
@olix0r You're asking good questions. What I would like the problem statement to convey is that that testing in Linkerd is currently not designed very well as it does provide a helpful UX, constrains the existing state (such as no existing Linkerd installation), and is not extensible for future tests such (such as testing multicluster).

What I do *not* think this RFC will change is tests on the individual level such that one can be ported as an example. I imagine that while we do not have a great option for programmatically setting cluster state, this will affect mostly the scripting side of test setup--essentially what happens when users run `bin/test-run`.

In the design phase those changes and the questions you listed can be further discussed and answered, but I'd like to know if the problem statement is at the right "level" of change you are expecting.

olix0r (approved):
Yes, problem statement looks fine. Let's get into the background/details!

PR #14, opened 2020-04-06
Design phase for CI Health Metrics

alpeb:
(followup to #11)

- Contribution Name: ci-health-metrics                                                                                 
- Implementation Owner: @alpeb, @kleimkuhler 
- Start Date: 2020-03-16                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            [5/220]
- Target Date: 2020-04-03                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  
- RFC PR: [linkerd/rfc#11](https://github.com/linkerd/rfc/pull/11)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
- Linkerd Issue:                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        [linkerd/linkerd2#4176](https://github.com/linkerd/linkerd2/issues/4176)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 
- Reviewers: @grampelberg                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  
                                                                                                                                                              

# Summary                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                        
[summary]: #summary                                                                                                                                           

This RFC proposes a service that would monitor the results from each CI run. It                                                                               
should allow us to catch consistently flaky tests.                                                                                                            
                                                                                                                                                              
# Problem Statement (Step 1)                                                                                                                                  
                                                                                                                                                              
CI jobs inevitably fail occasionally, either because of real problems with the                                                                                
code, or because of test flakiness. We'd like to stay informed about what tests                                                                               
are being persistently flaky, without having to dig into the CI run results                                                                                   
themselves. This will allow us to properly prioritize work to fix those                                                                                       
failures.                                                                                                                                                     

More specifically we'd like to get, for a given time span, the list of error                                                                                  
messages with their frequency aggregated globally, per workflow and per job.                                                                                  
                                                                                                                                                              
The first problem to address will be to expose the CI jobs results, which                                                                                     
currently are only available through Github's UI. With access to that data in                                                                                 
the appropriate format, we'll then be able to determine how to persist and                                                                                    
present it to operators.                                                                                                                                      

# Design proposal (Step 2)                                                                                                                                    
                                                                                                                                                              
## Exposing the errors                                                                                                                                        

There are two ways to expose CI run failures: annotations and artifacts.                                                                                      

### Annotations                                                                                                                                               

Annotations are created whenever an error message from any of the CI steps is                                                                                 
sent to stdout/stderr using the appropriate format, described                                                                                                 
[here](https://help.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-error-message).                                           
Annotation messages are visible in the summary page of a CI run. Currently our                                                                                
error messages aren't following that format, and the summary page only shows                                                                                  
something like "Process completed with exit code 1." whenever a job fails.                                                                                    
Annotations are surfaced through Github's Checks API.                                                                                                         
                                                                                                                                                              
### Artifacts                                                                                                                                                 

Instead of conforming to Github's annotation format, we could use a more                                                                                      
standard format such as xUnit XML reports. That would imply building some                                                                                     
wrapper around the tests to accommodate that format, and the results would be                                                                                 
then uploaded to the CI run's artifact repository (which we're already                                                                                        
leveraging to store docker images). Then another tool can consume those reports.                                                                              
                                                                                                                                                              
But I haven't found such a tool that would consume xUnit reports for showing                                                                                  
historical trends, which is the main goal of this RFC (being able to catch                                                                                    
consistently flaky tests). For this reason I'm proposing we go down the                                                                                       
annotations route.                                                                                                                                            

(Although do raise your voice if you find an appropriate tool for xUnit                                                                                       
reports!)                                                                                                                                                     
                                                                                                                                                              
## Formatting                                                                                                                                                                                                                                                                                                                

We have integration tests and checks written in Go, Javascript and Bash. Each of                                                                              
those environments would require changing how they report errors to properly                                                                                  
format them as explained above.                                                                                                                               
                                                                                                                                                              
For example Go test failures would need to be modified like so:                                                                                               

```                                                                                                                                                           
go fail(t, "linkerd upgrade config command failed", "linkerd upgrade config command failed\n%s\n%s", out, stderr)                                             
```                                                                                                                                                           

Where `fail` would output a string like                                                                                                                       

```                                                                                                                                                           
::error file=upgrade_test.go,line=103::linkerd upgrade config command failed                                                                                  
```                                                                                                                                                           
that is interpreted by Github as an annotation. That function would also make                                                                                 
the original `t.Fail("linkerd upgrade config command failed\n%s\n%s", out,                                                                                    
stderr)` call to report the to Go's testing framework.                                                                                                        

Note that the annotation will just contain the generic error message (no                                                                                      
details) as we'd like to aggregate the number of appearances of such messages.                                                                                

## Available Data                                                                                                                                             

Once the error information is exposed, these are the pieces of data available                                                                                 
for each CI run:                                                                                                                                              

- Check Suite ID                                                                                                                                              
- Check Run ID                                                                                                                                                
- Workflow name                                                                                                                                               
- Job name                                                                                                                                                    
- Completion status                                                                                                                                           
- Start timestamp                                                                                                                                             
- Finish timestamp                                                                                                                                            
- Test/Check name                                                                                                                                             
- File name                                                                                                                                                   
- Error message                                                                                                                                               

Which are retrieved using the following Github API calls:                                                                                                     

```                                                                                                                                                           
# This gives us a list of check suite IDs:                                                                                                                    
GET /repos/linkerd/linkerd2/actions/runs                                                                                                                      

# For each check suite ID, this gives us the check run ID, job name, workflow                                                                                 
# ID, completion status and timestamps:                                                                                                                       
GET /repos/linkerd/linkerd2/check-suites/:check_suite_id/check-runs                                                                                           

# For each workflow ID, this gives us the workflow name                                                                                                       
# (this can be cached across check runs):                                                                                                                     
GET https://api.github.com/repos/linkerd/linkerd2/actions/workflows/:workflow_id                                                                              

# For each check run ID, we invoke the annotations API which gives us the file                                                                                
# name and error message                                                                                                                                      
GET repos/linkerd/linkerd2/check-runs/:check_run_id/annotations                                                                                               
```                                                                                                                                                           

## New Workflow for Reports                                                                                                                                   

After each CI run finishes, a new Github Actions event would be dispatched that                                                                               
should be caught by a new, separate workflow whose sole responsibility is to                                                                                  
build a report.                                                                                                                                               

That workflow will first fetch the data described above. While data is available                                                                              
for approximately 3 months back, we're only interested in the data for the last                                                                               
  month. We have the option to fetch and process all the data upon each workflow                                                                              
  run, which will alleviate us from dealing with any kind of state, or                                                                                        
  alternatively, store it in some cloud DB service. As a first approach, and                                                                                  
  given the small amount of data we have to deal with, we can do the former and                                                                               
  see how that performs.                                                                                                                                      

## Reports                                                                                                                                                    

The reports would consist in tables and their corresponding charts for the                                                                                    
following data:                                                                                                                                               

- Success rate                                                                                                                                                
- Error frequency (showing most frequent error messages)                                                                                                      
- Duration                                                                                                                                                    

We can have one set of tables/charts for last month's data and another for the                                                                                
last week's data. Also, we can aggregate per workflow and per job, in the same                                                                                
chart.                                                                                                                                                        

The final report can be an HTML page containing the tables and pointing to chart                                                                              
files generated  with a library like [chart](https://github.com/vdobler/chart),                                                                               
all stored as workflow artifacts.                                                                                                                             

# Prior art                                                                                                                                                   

[prior-art]: #prior-art  

Github's Checks API have only until recently become beta, so it's                                                             
expected there aren't well known tools that leverage that information.

I did find the [BuildPulse](https://github.com/marketplace/buildpulse/) Github
App, that specializes on tracking flaky tests in Github Actions. I couldn't make
it work in a Linkerd2 fork. And presumably it will also require the changes
described in Phase 1 above for it to properly give us actionable data instead of
data about the generic error messages we currently have. BuildPulse is closed
source and a paid service.

# Future possibilities

[future-possibilities]: #future-possibilities

Once we expose the CI metrics, we open the door to present this data however we
wish in the future.

Reports will be stored as artifacts for the new workflow. For ease of
consumption, we can think of publishing them under some section of the website.


design/0002-ci-health-metrics.md
```diff
@@ -0,0 +1,154 @@
...
+consistently flaky tests). For this reason I'm proposing we go down the
+annotations route.
+
+(Although do raise your voice if you find an appropriate tool for xUnit
+reports!)
+
+## Formatting
+
+We have integration tests and checks written in go, javascript and bash. Each of
+those environments would require a wrapper around their output to properly
```
adleong: Can you talk a bit more about the wrapper?  What events from the original output are turned into annotations?  Tests that pass?  Tests that fail?  Error messages?  Stack traces?  Log output?  How would it identify which lines of the original output need to be turned into annotations?  Regular expressions?

design/0002-ci-health-metrics.md
```diff
@@ -0,0 +1,154 @@
...
+format errors as explained above. The error message would need to have at least
+two elements: A generic message that can be aggregated in a counter (see below)
```
adleong: It seems like it would be useful to have more structured fields in the message including things like:
Test name, source file, error message, line number, test duration

design/0002-ci-health-metrics.md
```diff
@@ -0,0 +1,154 @@
...
+and a message with more details about this particular run.
+
+## Architecture
+
+The plan being to ultimately present the historical trend of errors, the obvious
```
adleong: To me, just a regular (not timeseries) database seems like a better fit here.  It lets us have columns for things like test name, source file, error message, line number, test duration, timestamp, etc. without needing to worry about making all of these fields into Prometheus labels and handling the cardinality.

This easily lets us do queries like:
* list all of the tests sorted by success rate over the last 30 days
* for a given test, list the most recent 10 failures and their error messages

This feels a lot more like a database of events and less like a timeseries.

design/0002-ci-health-metrics.md
```diff
@@ -0,0 +1,154 @@
...
+underlying data store would be a time series db.
+
+A familiar architecture could then be a Prometheus exporter fed off of data
+gathered from the Github API, collected by Prometheus and displayed by Grafana.
+
+### Prometheus exporter
+
+A process would poll the Github Actions and Checks API and expose a `/metrics`
```
adleong: If we use a traditional database instead, this process would simply write rows into the DB rather than exposing a metrics endpoint.

design/0002-ci-health-metrics.md
```diff
@@ -0,0 +1,154 @@
...
+number of occurrences for each error message during the selected time span.
+
+These graphs could be aggregated per workflow, job and step.
+
+### Back-filling
+
+In order to start seeing results right away we need to back-fill Prometheus TSDB
+using all the results returned from the first API call referred above. Using
+[this approach](https://github.com/jjneely/stuff/tree/master/cmd/historygen)
+with some modifications I've been able to successfully achieve that.
```
adleong: In order to do back-filling, we would need to generate annotations from the log output of previous runs, right?  How would that work?

design/0002-ci-health-metrics.md
adleong: Either way, it would be helpful to list the columns or labels that we plan to record.

design/0002-ci-health-metrics.md
alpeb: The Github API calls I described above will return all the historical CI runs data from the past few months (right now I don't know exactly how far back)

design/0002-ci-health-metrics.md
alpeb: The sole fact these are data points collected over time, led me to suggest a TSDB. I wasn't worried by cardinality given the few amount of data this will handle. I'll have to confess though I don't know how easy it'd be to manage that data as opposed to simple SQL queries in a regular database, so I'm willing to reconsider.

design/0002-ci-health-metrics.md
alpeb: We could indeed capture more information. Note though that the specific problem we want to address is catching flaky tests.

design/0002-ci-health-metrics.md
alpeb: I'll add more details. In the meantime, to clarify, the plan is to only turn error messages into annotations, to build a counter that tracks errors that keep on happening. For example in go we'd fail with something like
```
t.Fail(format("linkerd upgrade config command failed", "linkerd upgrade config command failed\n%s\n%s", out, stderr)
```
Where `format` would return a string like `"::error::MSG"` (Github actions annotation format) where MSG could be a json array with the generic message ("linkerd upgrade config command failed") and the specific message ("linkerd upgrade config command failed\n%s\n%s", out, err). I'm not sure though on the necessity of the specific message. Other metadata like workflow/job/step can be collected in the Github API call.

design/0002-ci-health-metrics.md
adleong: If we plan to put the detailed error messages as Prometheus labels, I would assume these are all unique and we would get each data point alone in its own timeseries, which is kind of a weird way to use Prometheus.

design/0002-ci-health-metrics.md
adleong: Ah, but we'd only be able to backfill back to the point when we started formatting error messages as annotations, right?

design/0002-ci-health-metrics.md
adleong: Good point.  Additional information could be helpful for taking action once a flaky test has been discovered though.

design/0002-ci-health-metrics.md
adleong: Ah, so this involves a code change to our tests.  Understood.

design/0002-ci-health-metrics.md
alpeb: True, I missed that detail! Back-filling would only capture the generic "Process completed with exit code 1." messages, so this won't be worth it after all...

alpeb:
Thanks for the review @adleong. You convinced me Prometheus isn't the right match here. I'll push a new version based on this feedback :+1:

alpeb:
I've updated the proposal with a simpler approach on how to generate these reports. The markup in this PR's description has been updated accordingly.

grampelberg (approved):
I love where this got to! Lazy Thomas would just shove this all into a cloud DB and do SQL queries for the report but that's a minor optimization that might make sense once implementation starts. Ship it!

adleong (approved):
Love this.  Making it stateless as a first pass seems great!

admc (approved):
👍

admc:
Awesome!

PR #15, opened 2020-04-15
Add source metadata RFC

adleong:
When a meshed pod sends outgoing HTTP requests the Linkerd proxy records metrics
such as latency and request counters and scopes those metrics by the
destination.  This is done by setting a `dst_X` label on the Prometheus metrics
where `X` is the destination workload kind.  On the other hand, when a meshed
pod receives incoming HTTP requests, there is no equivalent scoping of the
metrics by source.  In other words, there is no `src_X` Prometheus label, making
it impossible to break down metrics for incoming traffic by source.  This RFC
proposes adding `src_X` Prometheus labels for incoming HTTP traffic.

### DIFF
--- a/design/0003-source-metadata.md
+++ b/design/0003-source-metadata.md
@@ -37,6 +37,25 @@ does this make the query awkward, it also means that resource-to-resource
 metrics can only be observed on the client side, never on the server side.  This
 limits our ability to measure network latency.
 
+More specifically, here are some examples of questions that cannot be answered
+without introducing source metadata:
+
+* We cannot compare client-side and server-side metrics for the same traffic to
+  identify latency or errors introduced between the two Linkerd proxies (e.g. by
+  the network or by other intermediary proxies)
+* It is difficult to present traffic metrics in a consistent way: top line
+  resource metrics are measured on the server-side by default, but in order to
+  view a breakdown of these metrics by source, the absesnse of source metadata
+  means that we have to switch to displaying client-side metrics.  This is
+  confusing at best and misleading at worst.
+* For traffic from unmeshed sources, the problem is even worse.  In this case
+  we don't have client-side metrics at all and can't display any metrics for
+  this traffic.  Introducing source metadata would allow us to distinguish
+  between meshed and unmeshed traffic on the server side.  While we would not
+  be able to distinguish between different unmeshed sources, we would at least
+  be able to show metrics for traffic from all unmeshed sources aggregated
+  together.
+
 Adding source metadata to HTTP traffic metrics would enable improvements in the
 Linkerd Grafana dashboard, 3rd party tools that consume Linkerd's Prometheus
 metrics, the controller's StatSummary API, and consequently the `linkerd stat`
--- a/design/0003-source-metadata.md
+++ b/design/0003-source-metadata.md
@@ -17,7 +17,7 @@ where `X` is the destination workload kind.  On the other hand, when a meshed
 pod receives incoming HTTP requests, there is no equivalent scoping of the
 metrics by source.  In other words, there is no `src_X` Prometheus label, making
 it impossible to break down metrics for incoming traffic by source.  This RFC
-proposes adding `src_X` Prometheus labels for incoming HTTP traffic.
+proposes adding `src_X` Prometheus labels all meshed HTTP traffic.
 
 # Problem Statement (Step 1)
 
@@ -49,7 +49,7 @@ this proposal.
 
 We will use the [Downward
 API](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#the-downward-api)
-to create and mound a volume in meshed pods which contains a file with the pod
+to create and mount a volume in meshed pods which contains a file with the pod
 owner's resource name and kind.  This is very similar to [the approach used to
 populate pod labels on span
 annotations](https://github.com/linkerd/linkerd2/pull/4199).  These values can
@@ -77,7 +77,6 @@ resource_kind="deployment"
 The proxy would read this file to get a list of labels to apply to its traffic
 metrics.  The way the proxy uses these labels differs for the inbound and
 outbound stacks.
-
 
 For the inbound stack, the proxy will prepend the `dst_` prefix to these labels
 and use them to scope the inbound metrics.  This will result in metrics such as:
@@ -111,16 +110,12 @@ request_total{
 }
 ```
 
-The outbound stack will also encode the source labels in an HTTP header of the
-outgoing request called `l5d-src-labels`.  An example of the value of this
-header would be: `resource_name=web,resource_kind=deployment`.  Encoding this
-source metadata on the request allows it to used by the inbound stack of the
-destination proxy.  Remember when we said we'd come back to inbound?
+The outbound stack will also share the source labels with the inbound stack of
+the destination proxy.  Remember when we said we'd come back to inbound?
 
-In addition to populating the `dst_` labels, the inbound stack will also read
-the `l5d-src-labels` HTTP header from the request, prepend `src_` to them, and
-add them to the label scope.  Thus, the complete inbound metrics would actually
-look like:
+In addition to populating the `dst_` labels, the inbound stack will also use the
+source labels, prepend `src_` to them, and add them to the label scope.
+Thus, the complete inbound metrics would actually look like:
 
 
 ```
@@ -132,6 +127,9 @@ request_total{
   src_resource_kind="deployment",
 }
 ```
+
+The details of how the source labels will be shared between the source outbound
+proxy and the destination inbound proxy are out-of-scope of this RFC.
 
 Note that all of the changes described here are additive to the existing
 Prometheus labels and would not introduce any backwards incompatibility.
@@ -162,17 +160,18 @@ proxy resolve IP addresses to resources would either require an API call which
 would introduce latency in the critical data path or would require an in-process
 cache which would increase the proxy memory footprint.
 
-There is also precedent in Linkerd for proxies to communicate metadata to one
-another using the `l5d-*` headers.  For example, the fully qualified authority
-is communicated between proxies using the `l5d-dst-canonical` header.
+In Istio configuration which do not use mixer, source context is communicated
+through a base64 encoded map of key-value pairs which in an HTTP header.  This
+approach requires reading and decoding this header on a per-request basis, even
+though we know that the source metadata will be the same for all requests on a
+single connection.  By communicating source metadata at the connection level,
+we can avoid doing this work for each request.
 
 # Unresolved questions
 
 [unresolved-questions]: #unresolved-questions
 
-Existing metrics stack modules in the proxy have a fixed label scope per stack.
-However, this proposal describes labels set dynamically from the incoming
-request. Is this feasible in the proxy stack architecture?
+The details of how to share source metadata will be discussed in another RFC.
 
 # Future possibilities
 
--- a/design/0003-source-metadata.md
+++ b/design/0003-source-metadata.md
@@ -10,54 +10,90 @@
 
 [summary]: #summary
 
-When a meshed pod sends outgoing HTTP requests the Linkerd proxy records metrics
-such as latency and request counters and scopes those metrics by the
-destination.  This is done by setting a `dst_X` label on the Prometheus metrics
-where `X` is the destination workload kind.  On the other hand, when a meshed
-pod receives incoming HTTP requests, there is no equivalent scoping of the
-metrics by source.  In other words, there is no `src_X` Prometheus label, making
-it impossible to break down metrics for incoming traffic by source.  This RFC
-proposes adding `src_X` Prometheus labels all meshed HTTP traffic.
+Linkerd's metrics API lacks the ability to query for server-side metrics when
+doing a resource-to-resource query.  When metrics for a single resource are
+requested, the returned metrics are always measured on the server-side.  But
+when metrics for traffic between two resources is requested, the returned
+metrics are always measured on the client-side.  This behavior is both
+unintuitive and limiting.  Users are frequently unaware or surprised that these
+two types of queries are measured differently.  Without any way to measure
+resource-to-resource traffic metrics on the server-side, it is impossible to
+compare client-side to server-side metrics for this type of traffic to identify
+network introduced latency or errors.
 
 # Problem Statement (Step 1)
 
 [problem-statement]: #problem-statement
 
-Linkerd is not able to add `src_X` labels today because it simply has no
-knowledge of the source resource.  It knows the peer socket address of the
-source, but has no mechanism to convert that address into a Kubernetes resource
-name or type.  This is in contrast to the `dst_X` metadata which the proxy
-gets from the Destination controller when doing service discovery look-ups.
+Linkerd's metrics API requests have this structure:
 
-This asymmetry in metadata can be very limiting when doing queries.  It is
-impossible to determine who the clients of a resource are by looking at that
-resource's metrics alone.  Instead, we need to query the outbound metrics of all
-other resource to find a client with the appropriate `dst_X` label.  Not only
-does this make the query awkward, it also means that resource-to-resource
-metrics can only be observed on the client side, never on the server side.  This
-limits our ability to measure network latency.
+```
+message StatSummaryRequest {
+  ResourceSelection selector = 1;
+  string time_window = 2;
 
-More specifically, here are some examples of questions that cannot be answered
-without introducing source metadata:
+  oneof outbound {
+    Empty none = 3;
+    Resource to_resource   = 4;
+    Resource from_resource = 5;
+  }
 
-* We cannot compare client-side and server-side metrics for the same traffic to
-  identify latency or errors introduced between the two Linkerd proxies (e.g. by
-  the network or by other intermediary proxies)
-* It is difficult to present traffic metrics in a consistent way: top line
-  resource metrics are measured on the server-side by default, but in order to
-  view a breakdown of these metrics by source, the absesnse of source metadata
-  means that we have to switch to displaying client-side metrics.  This is
-  confusing at best and misleading at worst.
+  bool skip_stats = 6;  // true if we want to skip stats from Prometheus
+  bool tcp_stats = 7;
+}
+```
 
-Adding source metadata to HTTP traffic metrics would enable improvements in the
-Linkerd Grafana dashboard, 3rd party tools that consume Linkerd's Prometheus
-metrics, the controller's StatSummary API, and consequently the `linkerd stat`
-CLI command and Linkerd dashboard.  These improvements are out of the scope of
-this proposal.
+If the `outbound` field is set to `none` then metrics are measured on the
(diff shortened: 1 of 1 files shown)


siggy:
To confirm my own understanding, this proposes to modify all request/response metrics exported via the linkerd-proxy's `/metrics` endpoint to include the following label keys:
- `direction`
- `src_workload_kind`
- `src_workload_name`
- `dst_workload_kind`
- `dst_workload_name`

This implies three changes to the proxy `/metrics` endpoint:
- adding `src_*` labels to inbound request/response metrics
- adding `dst_*` labels to inbound request/response metrics
- adding `src_*` labels to outbound request/response metrics

Context: linkerd/linkerd2#4101, linkerd/linkerd2#4102

olix0r:
* Are src labels added to outbound metrics?
* Are dst labels added to inbound metrics?

@siggy note that it won't be _all_ metrics. it will be metrics for all meshed communication. (We can't have dst metrics for services that bypass discovery).

design/0003-source-metadata.md
```diff
@@ -0,0 +1,182 @@
...
+[summary]: #summary
+
+When a meshed pod sends outgoing HTTP requests the Linkerd proxy records metrics
+such as latency and request counters and scopes those metrics by the
+destination.  This is done by setting a `dst_X` label on the Prometheus metrics
+where `X` is the destination workload kind.  On the other hand, when a meshed
+pod receives incoming HTTP requests, there is no equivalent scoping of the
+metrics by source.  In other words, there is no `src_X` Prometheus label, making
+it impossible to break down metrics for incoming traffic by source.  This RFC
+proposes adding `src_X` Prometheus labels for incoming HTTP traffic.
```
siggy: Assuming this proposal _also_ includes:

- adding `dst_X` labels for incoming traffic
- adding `src_X` labels for outgoing traffic

... I'd like to see that called out here.

design/0003-source-metadata.md
```diff
@@ -0,0 +1,182 @@
...
+CLI command and Linkerd dashboard.  These improvements are out of the scope of
+this proposal.
+
+# Design proposal (Step 2)
+
+[design-proposal]: #design-proposal
+
+We will use the [Downward
+API](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#the-downward-api)
+to create and mound a volume in meshed pods which contains a file with the pod
```
siggy: mount ;)

design/0003-source-metadata.md
```diff
@@ -0,0 +1,182 @@
...
+pod-template-hash="5cb99f85d8"
+```
+
+Our startup script would rewrite this to keep only the labels that we want and
+to put them in the desired format:
+
+```
+resource_name="web"
+resource_kind="deployment"
+```
```
siggy, line 123: Prefer `workload_name` and `workload_kind`, which conform more closely to k8s' notion of workloads:
https://kubernetes.io/docs/concepts/workloads/controllers/deployment/

A `resource` feels like a broader concept. A resource could be a node, for example. We could be even more specific and use `controller`, but I think that caries other connotations, and also would imply we supported `ReplicaSet`.

design/0003-source-metadata.md
```diff
@@ -0,0 +1,182 @@
...
+  direction="inbound",
+  dst_resource_name="web",
+  dst_resource_kind="deployment",
+  src_resource_name="vote-bot",
+  src_resource_kind="deployment",
+}
+```
+
+Note that all of the changes described here are additive to the existing
+Prometheus labels and would not introduce any backwards incompatibility.
```
siggy, line 184: 👍

siggy (commented):
A few comments but overall this is awesome! My reading of this is that it will solve both linkerd/linkerd2#4101 and linkerd/linkerd2#4102.

design/0003-source-metadata.md
```diff
@@ -0,0 +1,182 @@
...
+- Target Date: 
+- RFC PR: 
+- Linkerd Issue: 
+- Reviewers: 
+
+# Summary
+
+[summary]: #summary
+
+When a meshed pod sends outgoing HTTP requests the Linkerd proxy records metrics
```
grampelberg: I have foolishly not looked into the implementation enough for this. So, let me get a better understanding of how this works today.

## Source

- `dst_X` label is added by the proxy
- `direction=outbound` is added by the proxy
- `X` label is added by prometheus

## Destination

- `X` label is added by prometheus
- `direction=inbound` is added by the proxy

Am I correct there?

design/0003-source-metadata.md
grampelberg: @siggy oooh, these are really great points, it actually highlights that we want:

```
resource_name="web-123-123456"
resource_kind="pod"
owner_name="web"
owner_kind="deployment"
```

Having `resource_kind=node` actually makes a ton of sense to me (though I have no idea how we'd make it actually work).

design/0003-source-metadata.md
```diff
@@ -0,0 +1,182 @@
...
+Linkerd Grafana dashboard, 3rd party tools that consume Linkerd's Prometheus
+metrics, the controller's StatSummary API, and consequently the `linkerd stat`
+CLI command and Linkerd dashboard.  These improvements are out of the scope of
+this proposal.
+
+# Design proposal (Step 2)
+
+[design-proposal]: #design-proposal
+
+We will use the [Downward
```
grampelberg, line 98: Is there a reason to not just use env vars instead?

design/0003-source-metadata.md
```diff
@@ -0,0 +1,182 @@
...
+annotations](https://github.com/linkerd/linkerd2/pull/4199).  These values can
+be extracted from pod labels such as `linkerd.io/proxy-deployment`,
+`linkerd.io/proxy-daemonset`, etc.  We can create a simple startup script in the
+proxy container to preprocess these pod labels into the labels that we would
+like the proxy to use.  For example, the Downward API would create a volume with
+a file containing:
+
+```
+app="web-svc"
+linkerd.io/control-plane-ns="linkerd"
```
grampelberg, line 112: It's a bummer we can't get owner references through the API specifically and will need to do it with labels.

adleong:
@siggy @olix0r Yes, both inbound and outbound will have both dst and src metrics.  For inbound, the dst_X labels will be set to the pod's own workload (ie the labels loaded from the local file) and for outbound the src_X labels will be set to the pod's own workload.

I tried to make this explicit in the text of the design proposal and in the examples contained therein.  If you have any suggestions for how I can make this more clear, please let me know.

design/0003-source-metadata.md
adleong: As far as I can tell, Kubernetes is not super consistent about this terminology.  Pods are workloads according to the hierarchy in the left nav: https://kubernetes.io/docs/concepts/workloads/

The most technically correct is probably "controller", or more specifically, "pod controller", but we've never used this terminology like that in Linkerd before and I fear it can be easy confused with the "control plane".  The term "resource" is perhaps overly broad, but I think using it here like this would be mostly consistent with how it's used elsewhere in Linkerd.

Just my 20m.  I'd be perfectly fine with either resource or workload.

design/0003-source-metadata.md
```diff
@@ -0,0 +1,182 @@
...
+resource_name="web"
+resource_kind="deployment"
+```
+
+The proxy would read this file to get a list of labels to apply to its traffic
+metrics.  The way the proxy uses these labels differs for the inbound and
+outbound stacks.
+
+
+For the inbound stack, the proxy will prepend the `dst_` prefix to these labels
```
grampelberg, line 129: It sounded like @siggy found a prometheus config that could do this for us. I'm not sure it is worth it, with @Pothulapati working on BYO prometheus removing configuration from the scrape config feels like a net positive. Just wanted to make sure we'd weighed the tradeoffs.

design/0003-source-metadata.md
```diff
@@ -0,0 +1,182 @@
...
+  direction="outbound",
+  src_resource_name="web",
+  src_resource_kind="deployment",
+  dst_resource_name="emoji",
+  dst_resource_kind="deployment",
+  ...
+}
+```
+
+The outbound stack will also encode the source labels in an HTTP header of the
```
grampelberg: This raises a couple questions:

1. Do we care that this is an HTTP only solution? Seems like there'd be some value in having this for TCP as well.
1. Do we mind modifying the request?
1. What does this look like for inbound, non-meshed traffic?
1. Do care about malicious sources being able to pretend their metrics are from elsewhere?

design/0003-source-metadata.md
```diff
@@ -0,0 +1,182 @@
...
+introducing significant complexity.  The fact that Prometheus is able to scrape
+Linkerd proxies directly is a great property.  Alternatively, having the Linkerd
+proxy resolve IP addresses to resources would either require an API call which
+would introduce latency in the critical data path or would require an in-process
+cache which would increase the proxy memory footprint.
+
+There is also precedent in Linkerd for proxies to communicate metadata to one
+another using the `l5d-*` headers.  For example, the fully qualified authority
+is communicated between proxies using the `l5d-dst-canonical` header.
+
```
grampelberg, line 216: The non-mixer istio actually does this differently still. Have you taken a peek at how they're doing it now?

design/0003-source-metadata.md
adleong: That's correct.

design/0003-source-metadata.md
adleong: We could certainly do it either way.  Two reasons why I think a volume is appealing:

1. We have to do some munging of the labels and values to select the ones we want and get it into the format we want.  Feels cleaner to do that with files rather than in the environment.
2. There's precedent for getting labels into the proxy with files like this: https://github.com/linkerd/linkerd2/pull/4199

design/0003-source-metadata.md
adleong: Yeah.  :(

design/0003-source-metadata.md
adleong: Yes.  The inbound src_X labels and the outbound dst_X labels cannot come from the scrape config and must be provided by the proxy.  The inbound dst_X labels and outbound src_X labels *could* be added by the Prometheus scrape config as @siggy describes but it feels more consistent to have the proxy just provide all of the labels.

design/0003-source-metadata.md
adleong: 1. This would definitely be useful for TCP as well.  Doing so would require communicating the source metadata on the connection somehow.  I'd call this out-of-scope for now, but interesting future work.
2. We already set some Linkerd specific l5d-* information headers on proxy to proxy requests.
3. Such traffic could simply be missing the `src_X` labels.  Or, if it's preferable, these labels could be set to a blank or placeholder value.  Open to suggestions here.
4. This is a good point.  We should only respect these headers when the source has a validated identity.

design/0003-source-metadata.md
adleong: I haven't.  I'll check it out.

design/0003-source-metadata.md
grampelberg: 👍 that sounds like a great set of reasons to me =)

design/0003-source-metadata.md
olix0r: I'm onboard with most of this doc until we get to this detail. I'd very much prefer to do this at the connection-level: in addition to being able to support TCP metrics, this would substantially reduce the overhead of the feature from a runtime-cost (in terms of latency, cpu) point of view. Because this metadata can only be transferred within the mesh, our HTTP/2 multiplexing further reduces any need to rebroadcast this information.

I think that this RFC should depend on a means to share connection-level context between proxies and not get into those weeds (i.e. let's call it out scope, but blocking implementation). And then I think we ought to get an RFC scoped tightly around that context sharing and how to implement it. Happy to help as needed.

design/0003-source-metadata.md
```diff
@@ -0,0 +1,181 @@
...
+pod receives incoming HTTP requests, there is no equivalent scoping of the
+metrics by source.  In other words, there is no `src_X` Prometheus label, making
+it impossible to break down metrics for incoming traffic by source.  This RFC
+proposes adding `src_X` Prometheus labels all meshed HTTP traffic.
+
+# Problem Statement (Step 1)
+
+[problem-statement]: #problem-statement
+
+Linkerd is not able to add `src_X` labels today because it simply has no
```
olix0r: So, I understand this limitation. But can we be clearer about the _problem_ here?  What specific features / use cases are we not able to satisfy without these metrics?

_Especially_ because this proposal only includes this metadata on meshed connections (so we're guaranteed to have the client metrics).

What do these labels enable, concretely, that we can't do otherwise? Or if we do not add them, what risks do we incur?

design/0003-source-metadata.md
```diff
@@ -0,0 +1,181 @@
...
+name or type.  This is in contrast to the `dst_X` metadata which the proxy
+gets from the Destination controller when doing service discovery look-ups.
+
+This asymmetry in metadata can be very limiting when doing queries.  It is
+impossible to determine who the clients of a resource are by looking at that
+resource's metrics alone.  Instead, we need to query the outbound metrics of all
+other resource to find a client with the appropriate `dst_X` label.  Not only
+does this make the query awkward, it also means that resource-to-resource
+metrics can only be observed on the client side, never on the server side.  This
+limits our ability to measure network latency.
```
olix0r: So? Is the problem "this is awkward?" or are there fundamental limitations that we incur here?

I want to be careful about looking at this benefit versus the cost of the implementation...

design/0003-source-metadata.md
adleong: The next paragraph talks about the limitations we face from not having these metrics.

design/0003-source-metadata.md
adleong: I've added some explicit examples of questions that we can't answer without source metadata.

design/0003-source-metadata.md
```diff
@@ -0,0 +1,200 @@
...
+  view a breakdown of these metrics by source, the absesnse of source metadata
+  means that we have to switch to displaying client-side metrics.  This is
+  confusing at best and misleading at worst.
+* For traffic from unmeshed sources, the problem is even worse.  In this case
+  we don't have client-side metrics at all and can't display any metrics for
+  this traffic.  Introducing source metadata would allow us to distinguish
+  between meshed and unmeshed traffic on the server side.  While we would not
+  be able to distinguish between different unmeshed sources, we would at least
+  be able to show metrics for traffic from all unmeshed sources aggregated
+  together.
```
olix0r: I don't think that this RFC actually addresses this issue... We already have client IDs on server metrics.

design/0003-source-metadata.md
```diff
@@ -0,0 +1,200 @@
...
+without introducing source metadata:
+
+* We cannot compare client-side and server-side metrics for the same traffic to
+  identify latency or errors introduced between the two Linkerd proxies (e.g. by
+  the network or by other intermediary proxies)
+* It is difficult to present traffic metrics in a consistent way: top line
+  resource metrics are measured on the server-side by default, but in order to
+  view a breakdown of these metrics by source, the absesnse of source metadata
+  means that we have to switch to displaying client-side metrics.  This is
+  confusing at best and misleading at worst.
```
olix0r: This is still a bit squishy... It's difficult, but we do it. I.e. for consumers of our API, there's no overhead. Right?

design/0003-source-metadata.md
```diff
@@ -0,0 +1,200 @@
...
+does this make the query awkward, it also means that resource-to-resource
+metrics can only be observed on the client side, never on the server side.  This
+limits our ability to measure network latency.
+
+More specifically, here are some examples of questions that cannot be answered
+without introducing source metadata:
+
+* We cannot compare client-side and server-side metrics for the same traffic to
+  identify latency or errors introduced between the two Linkerd proxies (e.g. by
+  the network or by other intermediary proxies)
```
olix0r: To put a finer point on this... there's a specific API call for which we return incorrect data, right? What is that API Call?

I think this may be the only problem we're trying to address with this proposal. I'd try to take out much of the other context about labels and asymmetry and focus very concretely on this issue.

design/0003-source-metadata.md
adleong: We do not present traffic metrics in a consistent way today.  We're forced to either hide the distinction between server-side and client-side metrics, which is deceitful, or be explicit that certain queries are measured on one side while other queries are measured on the other, which is confusing.  Today we split the difference by being vague.  So I see this a problem for consumers of the API.

design/0003-source-metadata.md
adleong: I don't think that's quite right.  I don't think any of our API calls return incorrect data, their behavior is just confusing and non-uniform.

design/0003-source-metadata.md
adleong: True, though we don't use that data for this purpose today.  We could use it, but it would be more straightforward to use source metadata.  While this would not impact consumers of the API, I still think that having a cleaner implementation is of some value.

However, I have removed this point for now.

design/0003-source-metadata.md
olix0r: This makes it sounds like its our metrics API that is the problem: that we have no way to distinguish these concepts at all.

To put it another way: if we made the changes proposed here, what impact would that have on our metrics API? Would we just subtly change the semantics of an existing call? Would we be unable to represent this call?

I view Linkerd's prometheus data as diagnostics... like log lines, basically. The format & details may change between releases as long as we satisfy our metrics API, which powers consumers within Linkerd. So, a problem statement phrased in terms of our prometheus data isn't all that compelling.

However, it absolutely _does_ sound like a problem if there are fundamental questions that our API cannot answer about mesh traffic. How do we improve Linkerd's metrics API so that it can properly differentiate client and serverside metrics? If this requires source labels, great, I'm all in.

adleong: Reworked the RFC to be API focused.

olix0r: Love the updated framing.

If it's alright with you, I'm going to suggest some further edits via PR to contextualize how I see this all fitting together.

olix0r:

siggy:
Noting that as we rethink our metrics and labeling, a nice goal would be to conform to OpenMetrics. One notable requirement is that all metrics with the same name (a MetricsFamily) have the same set of label names:
https://github.com/OpenObservability/OpenMetrics/blob/defac23507e1e78871f8de2648bd809427a1063d/specification/OpenMetrics.md#metric
https://github.com/prometheus/docs/blob/be5550b3759db4d134a9eea9b1a91de092dcc5a8/content/docs/instrumenting/writing_clientlibs.md#labels

Another goal should be to avoid missing metrics. If we start counting `requests_total{workload_name="foo"}`, we should initialize `failed_requests_total{workload_name="foo"}`:
https://github.com/prometheus/docs/blob/be5550b3759db4d134a9eea9b1a91de092dcc5a8/content/docs/practices/instrumentation.md#avoid-missing-metrics

More best practices:
https://prometheus.io/docs/instrumenting/writing_exporters/

sdhoward:
I'm just here to remind that people are still interested in this.

PR #16, opened 2020-04-15
Bring your own prometheus

Pothulapati:
Rendered: https://github.com/Pothulapati/rfc/blob/prom/design/0003-byop.md

# Bring your own prometheus

[summary]: #summary

This RFC aims to move prometheus as an add-on i.e not mandated but advised.
This includes moving the linkerd-prometheus manifests as a sub-chart and
allowing more configuration (more on this later), along with documentation
and tooling to use Linkerd with an existing prometheus installation.

# Problem Statement

Lots of users already have a exisiting prometheus installation in their
cluster and don't like managing another prometheus instance especially
because they could be resource intensive and painful. They usually
have a prometheus installation paired up with a long term monitroing
solutions like Cortex,InfluxDB, or third party solutions like Grafana
Cloud, etc. In those cases prometheus is just a temporary component, that
keeps pushing metrics into their long term solution, making users want to
re-use their exisitng prometheus.

The following are the list of problems that we plan to solve:

- Hard requirement of linkerd-prometheus.

- Configuring the current linkerd-prometheus for remote-writes, etc is
hard, because the prometheus configmap is overridden after upgrades.

- Not enough documentation and tooling for bringing your own prometheus
use-case.

Once the RFC is implemented, Users should not only be able to use their
exisiting prometheus, but configuration of the linkerd-prometheus for
remote-writes, etc should also be easy.

### DIFF
--- a/design/0003-byop.md
+++ b/design/0003-byop.md
@@ -162,8 +162,12 @@ have the relevant scrape-configs and default configuration.
 [prior-art]: #prior-art
 
 - [Prometheus  Operator](https://github.com/coreos/prometheus-operator) is a interesting tool,
- which can remove the problem of maintaining configuration for upgrades, etc from Linkerd, 
+ which can remove the problem of maintaining configuration for upgrades, etc from Linkerd,
  but comes with it's own controller and CRD's.
+- Specific to resource costs, We can consider things like tuning prometheus, dropping labels, etc making it
+  further lighter but there will still be some users who would want to use their exisitng prometheus. Irrespective
+  of this solution, we should definitely try to be light on linkerd-prometheus and also provide as many knobs for
+  configuration making it easy to manage. This could be addressed separately and [rfc/21](https://github.com/linkerd/rfc/pull/21) has some similarities.
 - Istio's [Bring your own Prometheus example](https://istiobyexample.dev/prometheus/). A similar helm
 based sub-chart model is used but this was with Mixer v1.
 - We already have docs for [exporting metrics](https://linkerd.io/2/tasks/exporting-metrics/).
--- a/design/0003-byop.md
+++ b/design/0003-byop.md
@@ -76,12 +76,18 @@ prometheus:
   enabled: true
   name: linkerd-prometheus
   image:
+  args:
+  # - log.format: json
   global:
     <global-prom-config> https://prometheus.io/docs/prometheus/latest/configuration/configuration/#configuration-file
   remote-write:
     <remote-write-config> https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write
-  alerts:
-    <alert-manager-config> https://prometheus.io/docs/prometheus/latest/configuration/configuration/#alertmanager_config 
+  alertManagers:
+    <alert-manager-config> https://prometheus.io/docs/prometheus/latest/configuration/configuration/#alertmanager_config
+  configMapMounts:
+  # - name: alerting-rules
+  #   subPath: alerting_rules.yml
+  #   configMap: linkerd-prometheus-rules
 ```
 
 All the current default configuration will be moved to `values.yaml` in the
--- a/design/0003-byop.md
+++ b/design/0003-byop.md
@@ -10,30 +10,37 @@
 
 [summary]: #summary
 
-This RFC aims to move prometheus as an add-on i.e not mandated
-but adviced. This includes changes to the linkerd-prometheus manifests to make it as a add-on and allowing more configuration (more on this later),
-along with documentation and tooling to use existing prometheus installation.
+This RFC aims to move prometheus as an add-on i.e not mandated but advised.
+This includes moving the linkerd-prometheus manifests as a sub-chart and
+allowing more configuration (more on this later), along with documentation
+and tooling to use Linkerd with an existing prometheus installation.
 
 # Problem Statement
 
 [problem-statement]: #problem-statement
 
-Lots of users already have a exisiting prometheus installation in their cluster and don't like managing
-another prometheus instance especially because they could be resource intensive and painful. They usually
-have a prometheus installation paired up with a long term monitroing solutions like Cortex,InfluxDB, or third party
-solutions like Grafana Cloud, etc.
-In those cases prometheus is just a temporary component, that keeps pushing metrics into their long term solution, making users want to re-use their exisitng prometheus.
+Lots of users already have a exisiting prometheus installation in their
+cluster and don't like managing another prometheus instance especially
+because they could be resource intensive and painful. They usually
+have a prometheus installation paired up with a long term monitroing
+solutions like Cortex,InfluxDB, or third party solutions like Grafana
+Cloud, etc. In those cases prometheus is just a temporary component, that
+keeps pushing metrics into their long term solution, making users want to
+re-use their exisitng prometheus.
 
-The following are the list of problems:
+The following are the list of problems that we plan to solve:
 
 - Hard requirement of linkerd-prometheus.
 
-- Configuring the current linkerd-prometheus for remote-writes, etc is hard, because the cm is overriden after upgrades, etc.
+- Configuring the current linkerd-prometheus for remote-writes, etc is
+hard, because the prometheus configmap is overridden after upgrades.
 
-- Not enough documentation for bringing your own prometheus case.
+- Not enough documentation and tooling for bringing your own prometheus
+use-case.
 
-Once the RFC is implemented, Apart from the default prometheus enabled installation Users should be
-able to use their exising prometheus instance, to also scrape metrics from the proxy
+Once the RFC is implemented, Users should not only be able to use their
+exisiting prometheus, but configuration of the linkerd-prometheus for
+remote-writes, etc should also be easy.
 
 # Design proposal
 
@@ -41,21 +48,25 @@ able to use their exising prometheus instance, to also scrape metrics from the p
 
 ## Prometheus as an Add-On
 
-By moving Prometheus as an add-on, We can make it as an optional component that can be used both through Helm and Linkerd CLI.
-The add-on model would well with upgrades, etc as per [#3955](https://github.com/linkerd/linkerd2/pull/3955). This is done by
-moving the prometheus components into a separate sub-chart in the linkerd repo
-and updating the code. The same health checks are also expected, but as a separate
-check Category, which would be transient i.e runs only when it recognises
-that the add-on is enabled.
+By moving Prometheus as an add-on, We can make it as an optional component
+that can be used both through Helm and Linkerd CLI. The add-on model would
+well with upgrades, etc as per [#3955](https://github.com/linkerd/linkerd2/pull/3955).
+This is done by moving the prometheus components into a separate sub-chart
+in the linkerd repo and updating the code. The same health checks are also
+expected, but as a separate check Category, which would be transient i.e
+runs only when it recognises that the add-on is enabled.
 
 By default, the prometheus add-on will be **enabled**.
 
 This also provides a single place of configuration for all things related to
-linkerd-prometheus. Though we are making linkerd-prometheus optional but having it is great for Linkerd and external users, for reasons like separation of
-concerns, etc. By allowing the `remote-write`, `alerts`, etc configuration in the
-linkerd-prometheus, we would decrease the need/want for external users to tie their existing prometheus to Linkerd,
+linkerd-prometheus. Though we are making linkerd-prometheus optional but
+having it is great for Linkerd and external users, for reasons like
+separation of concerns, etc. By allowing the `remote-write`, `alerts`, etc
+configuration in the linkerd-prometheus, we would decrease the need/want
+for external users to tie their existing prometheus to Linkerd,
 as it would be even easier for them to configure `remote-write`, etc
-configuration from their exisiting prometheus to the linkerd-prometheus right from the start.
+configuration from their exisiting prometheus to the linkerd-prometheus
+right from the start.
 
 The following default prometheus configuration is planned:
 
@@ -73,17 +84,24 @@ prometheus:
     <alert-manager-config> https://prometheus.io/docs/prometheus/latest/configuration/configuration/#alertmanager_config 
 ```
 
-All the current default configuration will be moved to `values.yaml` in the above format, allowing users to override this by using Helm or `--addon-config` through CLI.
+All the current default configuration will be moved to `values.yaml` in the
+above format, allowing users to override this by using Helm or  
+`--addon-config` through CLI.
 
 ## Using Exisiting Prometheus
 
-First, Documentation will be provided at `linkerd.io` to guide users on adding
-the `scrape-configs` required, to get the metrics from linkerd proxies, control plane components, etc. This can be found [here](https://github.com/linkerd/linkerd2/blob/master/charts/linkerd2/templates/prometheus.yaml#L26)
+First, Documentation will be provided at `linkerd.io` to guide users on
+adding the `scrape-configs` required, to get the metrics from linkerd
+proxies, control plane components, etc. This can be found
+[here](https://github.com/linkerd/linkerd2/blob/master/charts/linkerd2/templates/prometheus.yaml#L26)
 
-Once this is done, Users should be able to see golden metrics, etc in their exisiting prometheus instance.
-The `prometheus.url` field helps users in configuring Linkerd components i.e `public-api`,etc that needs prometheus access to the exisiting prometheus installation.
+Once this is done, Users should be able to see golden metrics, etc in their
+exisiting prometheus instance. The `prometheus.url` field helps users in
+configuring Linkerd components i.e `public-api`,etc that needs prometheus
+access to the exisiting prometheus installation.
 
 Flag interactions are defined below:
+
 | `prometheus.enabled` | `prometheus.url`              | Result                                                                                                                                                                                                                                           |
 |---------|------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | true    | ""               | Install linkerd-prometheus, and configure control-plane components to use it.                                                                                                                                                                    |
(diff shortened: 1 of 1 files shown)


design/0003-byop.md
```diff
@@ -0,0 +1,127 @@
...
+
+- Hard requirement of linkerd-prometheus.
+
+- Configuring the current linkerd-prometheus for remote-writes, etc is hard, because the cm is overriden after upgrades, etc.
+
+- Not enough documentation for bringing your own prometheus case.
+
+Once the RFC is implemented, Apart from the default prometheus enabled installation Users should be
+able to use their exising prometheus instance, to also scrape metrics from the proxy
+
```
grampelberg, line 45: Would you mind outlining the UX side of this a little? We want to make sure that none of the existing linkerd tooling breaks (instead using their prometheus for it all).

design/0003-byop.md
```diff
@@ -0,0 +1,127 @@
...
+# Design proposal
+
+[design-proposal]: #design-proposal
+
+## Prometheus as an Add-On
```
grampelberg: I would really like to see a discussion about how we're planning on using their prometheus from the linkerd codebase, what could go wrong and how we can make that connection less fragile.

design/0003-byop.md
```diff
@@ -0,0 +1,127 @@
...
+
+By moving Prometheus as an add-on, We can make it as an optional component that can be used both through Helm and Linkerd CLI.
+The add-on model would well with upgrades, etc as per [#3955](https://github.com/linkerd/linkerd2/pull/3955). This is done by
+moving the prometheus components into a separate sub-chart in the linkerd repo
+and updating the code. The same health checks are also expected, but as a separate
```
grampelberg: I think we're gonna need a different set of health checks for BYO prometheus.

design/0003-byop.md
```diff
@@ -0,0 +1,127 @@
...
+
+The following default prometheus configuration is planned:
+
+```yaml
+prometheus:
+  enabled: true
+  name: linkerd-prometheus
+  url: 
+  image: 
+  global:
```
grampelberg, line 81: Isn't it possible to load all these from files? Then folks could just manage configmaps and not actually change the chart in any way.

design/0003-byop.md
```diff
@@ -0,0 +1,127 @@
...
+    <remote-write-config> https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write
+  alerts:
+    <alert-manager-config> https://prometheus.io/docs/prometheus/latest/configuration/configuration/#alertmanager_config 
+```
+
+All the current default configuration will be moved to `values.yaml` in the above format, allowing users to override this by using Helm or `--addon-config` through CLI.
+
+## Using Exisiting Prometheus
+
+First, Documentation will be provided at `linkerd.io` to guide users on adding
```
grampelberg: I'm not sure it is in-scope for this, but it would be awesome if we could make the scrap config update automatically like we've done for the CLI args and help.

design/0003-byop.md
```diff
@@ -0,0 +1,127 @@
...
+
+# Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+- Should we support all prometheus configs in `linkerd-prometheus`?
+
+- How do we keep track of metrics that the proxies expose, to have consistent check tooling?
+
+- Most third party storage solutions have some auth mechanisms, should we allow that? IMO, we shouldn't as they can have a proxy, etc for cases like that.
```
grampelberg: This is a great one to call out. I think we should keep it out of scope for v1. As folks complain that we're missing support for their specific solution we can consider updating to handle that use case.

design/0003-byop.md
Pothulapati: Yep, Theres a section later in the doc about `Linkerd check tooling`. This was about having the current linkerd checks that we do for control-plane components for the prometheus add-on too.

design/0003-byop.md
alpeb: I think the first step is to extract our current Prometheus configmap into a separate yaml template (currently it lives together with the prom service and deployment). That configmap is currently being reworked in linkerd/linkerd2#4220, which adds the ability to plug-in recording and alerting rules through separate configmaps (no other prom configs items are pluggable like that from what I can tell).
Then this new Prometheus addon `values.yaml` could just have a few minimal entries like `enabled`, `image`, etc, and the new entries from that PR to configure rules. We could also have an entry to entirely override the provided Prometheus configmap with an external one, in which case the user would need to be advised to follow the scrape configs from the default configmap.
Makes sense?

design/0003-byop.md
```diff
@@ -0,0 +1,158 @@
...
+|---------|------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| true    | ""               | Install linkerd-prometheus, and configure control-plane components to use it.                                                                                                                                                                    |
+| false   | "prometheus.xyz" | No linkerd-prometheus, control-plane components will be configured to url                                                                                                                                                                        |
+| true    | "prometheus.xyz" | Install linkerd-prometheus, but stil configure control plane to use different url, useful when linkerd-prometheus is configured to remote-write into long term storage solution at url, Thus  powering dashboard and CLI through long term storage. |
```
alpeb, line 123: This is really interesting. Would it make sense to have some validation in this case ensuring that remote-write is setup?

design/0003-byop.md
```diff
@@ -0,0 +1,158 @@
...
+
+`--prometheus` flag can be added to linkerd check to run these tests.
+The following tests are planned:
+
+- Check if the required scrape-configs are present in proemetheus cm.
+( It should be noted that, the prometheus could be present in an differents
+namespace, etc)
+
+- Check if the common metrics are queriable in prometheus, by using the
+end-point i.e `prometheus.url` from the `linkerd-values` cm.
```
alpeb: A downside of trying to check the configmap is that we might not know which one it is, nor have access to it.

A downside of querying prometheus is that that might not give us enough info to verify the latest scrape config has been adopted (some subtle scrape config changes might not be reflected easily as a query result). I wonder if we could add some sort of version number in the scrape config that could be reflected as a query result...

design/0003-byop.md
```diff
@@ -0,0 +1,158 @@
...
+
+# Prior art
+
+[prior-art]: #prior-art
+
+- We already have docs for [exporting metrics](https://linkerd.io/2/tasks/exporting-metrics/).
+We could update those docs, with remote-write, etc.
```
alpeb: The intent of the Prior Art section is to gather experience from other projects about how they've trying solving a similar problem. For this particular RFC I skimmed what the Prometheus Operator project was doing to parametrize configuration, but that implied setting CRDs and what not, so that was too complex for what we need. Please let us know what you've researched elsewhere, and why/how it could be (or not) applied here.

design/0003-byop.md
Pothulapati: I added 
```
### User Expereince

Currently, `public-api`, `grafana` and `heartbeat` are the only three
components which are configured with a `prometheus-url` parameter,
All the Dashboards and CLI are powered through the `public-api`.
This allows us to swap prometheus without breaking any exisiting linkerd
tooling. Though it is expected that the `scrape-config` is configured
correctly, when this path is choosen.
```

Are you looking for anything more?

design/0003-byop.md
Pothulapati: yep, makes sense.

design/0003-byop.md
Pothulapati: From what I've spoke to @alpeb, This was definitely considered in-scope for the problem. We could definitely write some CLI tools for supporting, or even automatically updating the prom config with the current scrape-config, but having documentation on what the current scrape-config is, how to add it, etc should be good right?

design/0003-byop.md
Pothulapati: Yep, that make sense.Should we consider it in the hot path i.e to make the public-api perform the check when initializing or have Linkerd check tooling that checks this?

design/0003-byop.md
Pothulapati: > A downside of trying to check the configmap is that we might not know which one it is, nor have access to it.

Hmmm, we can mitigate this, by making the user pass the configmap name to the CLI, but there will still be the access problem but the good side should be that we can also have a command which tires to update the scrape config to the current version's scrape config.

> I wonder if we could add some sort of version number in the scrape config that could be reflected as a query result...

This is definitely interesting. but from what I searched, I couldn't find a way to add a param into the scrape-config but can we have a metric on the scrape-config version? If yes who would be the right candidate to check and add for the metric? Also, wondering if there would be any inconsistencies on the actual config, and this metric.

design/0003-byop.md
alpeb: That can be checked at install/upgrade time by adding a Helm validation like what we do in `indentity.yaml` for ensuring a cert is provided. I can't recall if that also gets applied during CLI install/upgrade; if not, then that'd be something to consider in the CLI scripts.

Pothulapati:
A POC of this rfc is available [here](https://github.com/Pothulapati/linkerd2/tree/final-prom), just to make sure there are no blockers.

design/0003-byop.md
```diff
@@ -0,0 +1,176 @@
+# Bring your own Prometheus
+
+- @pothulapati
+- Start Date: 2020-04-15
+- Target Date: 
+- RFC PR: [linkerd/rfc#0000](https://github.com/linkerd/rfc/pull/0000)
```
siggy: https://github.com/linkerd/rfc/pull/16 ;)

design/0003-byop.md
siggy: During implementation, I think we should audit all Prometheus queries from Linkerd, to ensure they are scoped to `job="linkerd-proxy"`, or similar. We don't want Linkerd accidentally querying for non-Linkerd metrics.

design/0003-byop.md
```diff
@@ -0,0 +1,176 @@
...
+above format, allowing users to override this by using Helm or  
+`--addon-config` through CLI.
+
+### Using Exisiting Prometheus
+
+First, Documentation will be provided at `linkerd.io` to guide users on
+adding the `scrape-configs` required, to get the metrics from linkerd
+proxies, control plane components, etc. This can be found
+[here](https://github.com/linkerd/linkerd2/blob/master/charts/linkerd2/templates/prometheus.yaml#L26).
+This can also be automated like we already do with Linkerd CLI help and args.
```
siggy: I agree some automation via CLI is useful. Some kind of CLI command that outputs the Prometheus ConfigMap necessary to scrape Linkerd proxies. We don't change that scrape config often, but we'll want to ensure that the ConfigMap provided to users matches their Linkerd installation.

Additionally, we should have some documentation on the storage requirements for Linkerd metrics in Prometheus. We workaround this today by having a 6h retention time, but this could become a significant issue for existing Prometheus installations with longer retention times. This could motivate customizations to the scrape config (like scrape interval changes), to help decrease storage requirements.

design/0003-byop.md
```diff
@@ -0,0 +1,176 @@
...
+but just log some warning. So, that users can still access things like TAP, Checks, etc through the dashboard and CLI.
+
+#### Linkerd CLI Tooling
+
+When users get their own Prometheus, It would be really great to have
+some tooling that allows users to check if prometheus is able to get the metrics.
+
+prometheus checks will be added to the default Linkerd health checks, which would check
+if the relevant metrics are present in prometheus by querying, and warn the users about the same. User's
+don't have to supply the `prometheusUrl` as we can access the same from `linkerd-values` Configmap.
```
siggy: 👍

siggy (commented):
Overall looks great!

design/0003-byop.md
Pothulapati: As @alpeb mentioned, I integrated stuff from https://github.com/linkerd/linkerd2/pull/4220 into this `prometheus` field. It dosen't help a lot, as only rules, etc can be added as files and other configuration like `remote_write`, `alerting` etc feels better to be taken directly.

> Then this new Prometheus addon values.yaml could just have a few minimal entries like enabled, image, etc, and the new entries from that PR to configure rules.

Users can always choose to ignore fields that he dont want to fill!

> We could also have an entry to entirely override the provided Prometheus configmap with an external one, in which case the user would need to be advised to follow the scrape configs from the default configmap.

This is interesting but it essentially feels like a BYOP case? and requires complication on maintaing the linkerd-addons configmap.

design/0003-byop.md
Pothulapati: Added this. On the additional documentation part, I definitely think we have some stuff in mind around retention times, etc, and more can be gathered as we understand problems when users use their existing prometheus.

design/0003-byop.md
Pothulapati: Hmm that makes sense, I will definitely see if there are any places where we use metrics that are scraped from stuff outside the scrape-config provided in documentation, and raise issues (if there are any)

alpeb (approved):
Thanks @Pothulapati , I think this covers all we have in mind, and should be enough to start implementation :+1:

grampelberg (approved):
Ship it!

design/0003-byop.md
```diff
@@ -0,0 +1,198 @@
...
+
+[problem-statement]: #problem-statement
+
+Lots of users already have a exisiting prometheus installation in their
+cluster and don't like managing another prometheus instance especially
+because they could be resource intensive and painful. They usually
+have a prometheus installation paired up with a long term monitroing
+solutions like Cortex,InfluxDB, or third party solutions like Grafana
+Cloud, etc. In those cases prometheus is just a temporary component, that
+keeps pushing metrics into their long term solution, making users want to
```
adleong, line 29: I'm not sure I really understand what users don't like about the existence of more than one Prometheus in their cluster.  What is the actual downside to having Linkerd-prometheus as a temporary storage for metrics before they are pulled into long-term storage via federation or remote-writes?  Does the linkerd-prometheus have a high resource cost?  Can we quantify this?  Is there some greater operational burden to having the linkerd-prometheus exist?  what is that burden?

design/0003-byop.md
```diff
@@ -0,0 +1,198 @@
...
+re-use their exisitng prometheus.
+
+The following are the list of problems that we plan to solve:
+
+- Hard requirement of linkerd-prometheus.
```
adleong, line 34: What's the actual problem that this solves, though?  Why is hard requirement of linkerd-prometheus bad?

design/0003-byop.md
```diff
@@ -0,0 +1,198 @@
...
+
+- Configuring the current linkerd-prometheus for remote-writes, etc is
```
adleong, line 36: :+1:

design/0003-byop.md
```diff
@@ -0,0 +1,198 @@
...
+hard, because the prometheus configmap is overridden after upgrades.
+
+- Not enough documentation and tooling for bringing your own prometheus
```
adleong, line 39: I think it's very important that we give a clear rubric in the documentation that allows users to figure out if BYO prometheus is something that they should do.  I think that for the (vast?) majority of users, even users that are already running a prometheus, using linkerd-prometheus in conjunction with their own prometheus is the right thing to do.

design/0003-byop.md
```diff
@@ -0,0 +1,198 @@
...
+- More checks can be added here, if there are some known
+configurations that don't go well with linkerd.
+
+*linkerd prom-config*
+
+A additional command can be provided in the CLI, which outputs the prometheus-config corresponding to that version of linkerd, which would
+have the relevant scrape-configs and default configuration.
+
+
+### Prior art
```
adleong, line 160: It would be helpful to describe alternative solutions that we considered and explored and why they were not suitable.  

For examples:
* agressive tuning of linkerd-prometheus
* exposing more tuning knobs of linkerd-prometheus
* dropping labels

design/0003-byop.md
Pothulapati: Yep, `linkerd-prometheus` isn't any different from the normal installation, and dosen't also get any specific tweaks for Linkerd. I definitely think users should be using the `linkerd-prometheus` as it dosen't do anything different and the operational burden shouldn't matter that much if there is already a Prometheus and long-term solution. We will be suggesting the same in the documentation.

On the resource cost,
We have the following in `ha` mode:
```
prometheusResources:
  cpu:
    limit: "4"
    request: 300m
  memory:
    limit: 8192Mi
    request: 300Mi
```

This isn't a lot but we have definitely seen cases where prometheus was being OOM, etc. I don't think this will be any better in the BYOP case, and the problem could be further amplified. 

I think, It's just one of those obv questions that users stumble upon initially "Can I use my existing prometheus?" but it would later make sense as they start using, to have separate one for Linkerd. 

IMO the main value of this rfc is separating out prometheus from the contorl-plane components in charts, configuration and having a life-cycle of its own for updating configuration, etc. BYOP is just another benefit that we are getting out of this. WDYT?

design/0003-byop.md
Pothulapati: Yep, Totally. and we will definitely have to highlight it in the documentation of BYOP case.

design/0003-byop.md
Pothulapati: This is a very good decision technically as there are clear separation of concerns here but a lot of people consider it as a shortcoming and have been asking the bringing their own prometheus use-case.

design/0003-byop.md
Pothulapati: > exposing more tuning knobs of linkerd-prometheus

This rfc address this by exposing the `global` field at `prometheus.global` where most tuning around retention period, etc can be configured. 

All the tuning stuff though being very helpful to make the linkerd-prometheus more stable, managable and not run into situations like OOM, etc when there is sudden spikes in queries, etc still won't address the user's question of getting their own prometheus, which we definitely heard a lot! (definitely not suggestible though)

I added this suggestion. Feel free to comment, if you would like me to add anything!

design/0003-byop.md
```diff
@@ -0,0 +1,202 @@
...
+[summary]: #summary
+
+This RFC aims to move prometheus as an add-on i.e not mandated but advised.
+This includes moving the linkerd-prometheus manifests as a sub-chart and
+allowing more configuration (more on this later), along with documentation
+and tooling to use Linkerd with an existing prometheus installation.
+
+## Problem Statement (Step 1)
+
+[problem-statement]: #problem-statement
```
adleong, line 21: I think this problem statement misses the point, at least as far as I understand it.

Having an existing prometheus installation is not, on its own, a good reason to omit linkerd-prometheus.  Furthermore, in the vast majority of cases, there is no significant resource or operational cost to running linkerd-prometheus.  This is true when linkerd-prometheus is used on its own for short-term retention and when linkerd-prometheus is paired with another solution for long-term storage.

As far as I'm aware the only good reason for bringing your own prometheus is if you operate at a large scale or have very specific prometheus configuration requirements and need to do extensive tuning or configuration of prometheus.

adleong (approved):
I understand the motivation and goals for this project and I think everyone involved does as well, but I want to call out that I don't think the problem statement in this RFC accurately reflects that.  I don't know if that should be a blocker to this moving forward or not.

The design plan looks great to me.

PR #18, opened 2020-04-16
problem: mTLS for all meshed communication

olix0r:
Provides a high-level scope for the `total-mesh-mtls` project. Design to
follow.


admc (approved):
This sounds great.

grampelberg (approved):
This looks great to me. I'm a little worried about throwing out the audit trail. We don't need it in the first version, but I'd hate to make some architectural decisions that have it become particularly difficult.

olix0r:
@grampelberg My main concern is that it raises questions about our use of Prometheus. Totally agree that it should follow. As I get into the design, I'll call it out as a future need.

PR #19, opened 2020-04-21
Add design phase for testing RFC

kleimkuhler:
[Rendered](https://github.com/kleimkuhler/rfc/blob/kleimkuhler/testing-design-phase/design/0003-isolated-integration-tests.md)

Signed-off-by: Kevin Leimkuhler <kevin@kleimkuhler.com>


olix0r:
@kleimkuhler I believe the last status was that you're doing some discovery/prototyping on the retry test? Are there parts of this that I should spend time on before you've done that?

kleimkuhler:
@olix0r Thomas and I talked about this yesterday and I am going to update this with a few additional thoughts. Overall this won't change that much, but I can add some explanation about why I am proposing certain changes. I would wait until I get that change in and I can ping you on here!

design/0003-isolated-integration-tests.md
```diff
@@ -0,0 +1,324 @@
...
+$ bin/deep-tests $PWD/target/cli/darwin/linkerd
+```
+
+### Constraints
+
+Tests will no longer be isolated by namespace. Each will require its own
+cluster so that it can require specific initial state.
+
+This will be the default behavior, but can be configured with a
+`--skip-cluster` flag that changes the behavior back to creating namespaces
```
alpeb, line 162: This flag is for skipping cluster creation?

alpeb (commented):
If I understand correctly, the plan is to have a separate script for each one of the integration tests (deep, upgrade, helm install, etc).
I was thinking it would be cool if this could be even more granular, say being able to run the integration tests just for `test/serviceprofiles`.
So a single entry point instead of multiple scripts might make more sense, that could receive as an argument the specific test to trigger, or the suite (deep, helm) to trigger. I haven't seen how to group tests into suites in go using the default machinery so maybe we'd have to introduce some tool for that.

design/0003-isolated-integration-tests.md
kleimkuhler: Yes and the behavior would change to what it currently is which is isolating by namespace.

design/0003-isolated-integration-tests.md
```diff
@@ -0,0 +1,324 @@
...
+bugs, but it does not provide a very positive UX experience to new or existing
+Linkerd developers.
+
+It is also unable to provide many options for testing complex cluster
+scenarios such as multi-cluster installations and cluster failure scenarios.
+
+The are several reasons for these existing problems and each should be
+addressed by the change this RFC is proposing.
+
+### UX
```
olix0r, line 40: I think it's also worth calling out the dangers of running tests with the wrong kube context.

design/0003-isolated-integration-tests.md
```diff
@@ -0,0 +1,324 @@
...
+After: Running all tests:
+
+```
+$ bin/tests $PWD/target/cli/darwin/linkerd
+```
+
+After: Running specific test:
+
+```
+$ bin/deep-tests $PWD/target/cli/darwin/linkerd
```
olix0r, line 153: I think this RFC should address the CLI argument. I never understood why we have to do this manually and it can't be done by the test harness...

kleimkuhler:
> I was thinking it would be cool if this could be even more granular, say being able to run the integration tests just for test/serviceprofiles

I don't think this should also script the Go test files. Currently, running the `test/` files is still just a `go test` command:

```
go test github.com/linkerd/linkerd2/test/serviceprofiles/... --linkerd $PWD/cli/darwin/linkerd --linkerd-namespace linkerd
```

I think it's important to keep the unit test interface still through `go`, and only script the integration tests. We can definitely add some more documentation about what flags you need to pass in, but I don't think we want to script them.

The point that I would want to make sure is clear is that there are two entry points to testing right now and I think that is okay. There are the unit tests in `test/` and they work like other `go test` files would. 

The second entry point is integration tests. The integration tests are not "native" to the project, they are just composed of several `test/` files. I think that's why they should stay as being scripted and should be where cluster management should take place.

design/0003-isolated-integration-tests.md
kleimkuhler: I usually have multiple CLI versions in my directory. If I'm testing something I just changed, then I would pass the `target/cli/..` binary. If I'm testing a version that I pulled with `docker-pull-binaries`, I would want to specify which version from `target/release/linkerd2-cli-..`.

I think the test harness handles this particularly well in that you can run the `test/` test files or integration tests with a specific version of the CLI.

If we moved this into the test harness, the developer would still need some way to specify that they want to run the tests with the CLI they just built, or a CLI that they just pulled from a PR or past release.

olix0r: I understand that it's useful to have the ability to use a custom CLI binary, but I think the vast majority of the time this is an unnecessary hurdle. I can't tell you how many times I've accidentally tested a stale binary because I was missing a rebuild...

I think about it like this... If I was running `go test ...`, it would handle building everything it needs to run the test. I don't have to run a bunch of `go build` commands before running tests.

How frequently do we actually want to test a specific, non-latest CLI version? Is it infrequent enough that we could support an optional `LINKERD_CLI` env to override it?

kleimkuhler: Yea I similarly do not run the Go on non-latest CLI versions that often. Most of my interaction with using a non-latest CLI version is pulling with `docker-pull-cli-binaries` to install and test some specific issue, or pulling the binaries from a PR and checking the behavior.

I would assume passing in the binary to the integration tests is there only because the Go tests require it. If passing in the binary was removed from the Go tests, I don't really think we'd need to keep it around for the integration tests.

I wouldn't want this change to mean that every time tests are run it has to rebuild the binary to make sure it's up to date because that can introduce a lot of time, but continuing to use a binary for testing definitely keeps this RFC scoped to it's original purpose.

If there is a way to avoid that, I would be okay removing the CLI argument from the Go tests as well as the integration tests.

olix0r: How does that binary typically get built today? It used to be built in Docker, which seems overblown. Or, even if there is a good reason to build it in Docker, we can start using newer Docker features to support build caching... But if the test script had a `go build` (or even `docker build` with caching) in it, I'd expect that to be very fast when you're not touching CLI code.

Basically, what I'm getting at is.. if I'm working on a given Service and want to run an integration test against it, what does the dev/test loop look like? Ideally, we just run a script with the test target...

olix0r, line 153: Also, I'm confused by the concept of `bin/deep-tests`... What's a "deep" test? What other kinds of tests are there? As a user, how do I know which script I want to run and why?

kleimkuhler: This is just the current [deep_integration_test](https://github.com/linkerd/linkerd2/blob/master/bin/_test-run.sh#L87-L94) that runs all the Go test files with no additional arguments.

There are integration tests such as uninstall or external issuer that would pass additional flags to the Go tests files, such as `--external-issuer=true`.

With the integration tests being separate scripts, the idea is that you would know what you want to run based off the name: `bin/external-issuer-test`, `bin/uninstall-test`, `bin/deep-test` (which maybe could be renamed to `all-test`/`all-default-test`)

olix0r: The more I think about it, I think the current setup isn't so bad... but it would helpful to include some examples of the full test loop including whatever commands are typicaly needed to build the CLI to be able to run tests

olix0r: What is it about these tests that vary, specifically?

Do they handle cluster setup/teardown different? Do they all end up invoking `go test`?

Also, we should consider a naming hierarchy that is tab-complete/sort friendly like `bin/test-all`.

Is there a way for this to be structured as, for instance `bin/test-harness ./path/to/issuer/tests`?

How does a developer understanding the mapping between the various test scripts and the actual source-code? Are the tests split into separate directories? If so, should we just keep a runner script in each test directory?

I don't really have enough context about the current state or how it's laid out to provide a lot in the way of guidance here, I guess. But, that probably makes me a decent advocate for an uninitiated contributor... Is the current setup documented? If I wanted to start adding a new type of tests, is there guidance in how I should approach it?

kleimkuhler: > Do they handle cluster setup/teardown different? Do they all end up invoking go test

They all end up invoking `go test` and passing in different flags specific to the test. For example, testing external issuer integration test will:
1. Run `install_test.go` (which is a test within itself for checking a clean install with the flags passed) with `--external-issuer=true` so that the setup installs a cluster with external issuer
2. Run `externalissuer/...` tests
3. Cleanup the namespace

The external issuer integration test takes care of this setup and teardown, but allows a developer to set some other specific state and run `go test github.com/linkerd/linkerd2/test/externalissuer/... [flags]`

> Also, we should consider a naming hierarchy that is tab-complete/sort friendly like bin/test-all

Sure that would make sense to do!

kleimkuhler: > Is there a way for this to be structured as, for instance bin/test-harness ./path/to/issuer/tests?

This is what `bin/test-external-issuer` would be. It setups the states, runs the test specified by the external issuer path, passes in the flags required (context, namespace, CLI) and then performs cleanup.

alpeb (approved):
Thanks for the explanations @kleimkuhler :+1: 

Point taken on calling `go test` directly to run specific tests under `/test`.  And it turns out most of this is already documented in `TEST.md`.

grampelberg (approved):
:shipit:

olix0r (approved):

admc:
Awesome!

PR #20, opened 2020-04-22
context sharing

adleong:
When one meshed pod talks to another, we would like for the source proxy to be
able to communicate additional information to the destination proxy.  This would
allow the destination proxy to expose better metrics and avoid duplicating work
already done by the source proxy.  This mechanism also gives us flexibility in
case further coordination between proxies is desired in the future.

### DIFF
--- a/problem/0004-context-sharing.md
+++ b/problem/0004-context-sharing.md
@@ -1,0 +1,60 @@
+- Contribution Name: context_sharing
+- Implementation Owner: Alex Leong
+- Start Date: 2020-04-21
+- Target Date:
+- RFC PR:
+- Linkerd Issue:
+- Reviewers:
+
+# Summary
+
+[summary]: #summary
+
+When one meshed pod talks to another, we would like for the source proxy to be
+able to communicate additional information to the destination proxy.  This would
+allow the destination proxy to expose better metrics and avoid duplicating work
+already done by the source proxy.  This mechanism also gives us flexibility in
+case further coordination between proxies is desired in the future.
+
+# Problem Statement (Step 1)
+
+[problem-statement]: #problem-statement
+
+When one meshed pod talks to another, communication passes from the application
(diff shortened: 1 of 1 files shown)


zaharidichev (commented):
This looks quite neat. What are some ideas around its implementation details

grampelberg (commented):
- WDYT about getting a list together of what we'd like to expose, at least as a v1? That should direct the design conversation.
- There should be a discussion in here around protocol support. My preference would be to have a solution that works at L4 instead of L7. That said, we should call out that it is a problem and then whether we're going to call it out of scope or not for this RFC.
- What are the implications here around multi-cluster?
- How does this fit in with the TCP mTLS work that is upcoming?
- Are there any questions we can answer for users by leveraging this data? I suspect that'll be intertwined with the list of metadata we'd like to expose.
- What's the implication for non-meshed clients and servers? Is there a world where we'd like to communicate with them as well? Maybe there's some interactions with identity?

adleong:
* I'd love to get more concrete about what we want to expose but I think that discussion should be attached to the specific feature that it enables.  In other words, we should discuss that in the "Source metrics" RFC and in the "Protocol Detection" RFC.  I see this RFC as providing a mechanism for side-channel communication between proxies that can be used to enable these other features.
* Added a note about protocol requirements
* Cross cluster requests are basically just regular proxy-to-proxy requests between the source and the gateway (and then between the gateway and the target).  So this mechanism will work for those individual hops.  To propagate information across multiple hops is a different conversation that starts to look like distributed tracing.
* From a problem statement point of view, this doesn't care about TCP mTLS at all.  From an implementation point of view, mTLS may give us a way to implement this as TLS extensions.  This is one of the options that should be considered in the design section.
* Yeah, I think this certainly enables a lot of good stuff from other RFCs.
* I think context communication with unmeshed peers is out of scope.

PR #21, opened 2020-04-24
Propose a stable metrics API

olix0r:
This is a rough sketch outlining some of the properties and constraints
of a stable Linkerd metrics API.

@adleong @grampelberg PTAL. As I thought more about #15, I realized that the problem we're solving is that we don't have any contract around our metrics interface, at all. We need one. I think when we look at this larger scope of problem, the concrete need for source-metrics becomes clear in order to satisfy a variety of requests in an access-controlled manner. But this is just one piece of the puzzle.

I think this implies that we move the Web/CLI off of the variety of use-case-specific metrics endpoints onto a generalized interface (where it makes sense, anyway).

### DIFF
--- a/problem/0000-metrics-v1.md
+++ b/problem/0000-metrics-v1.md
@@ -12,125 +12,21 @@
 
 [problem-statement]: #problem-statement
 
-Linkerd's `stat` API powers both the command-line and web interfaces, but it's tailored closely
-to their use cases and it has evolved only to express the features of these tools. Because of its
-narrow focus, integrators tend to read data directly from Prometheus, which features an
-expressive query language and a variety of external integrations/extensions. This presents an
-implicit, undocumented relationship between the proxy's metrics exposition structure and traffic
-in Kubernetes.
+Linkerd's `stat` API powers both the command-line and web interfaces, but
+it's tailored closely to their use cases and it has evolved only to express
+the features of these tools. Therefore, anyone trying to access this data
+must access Linkerd's Prometheus API directly. This presents several
+problems:
 
-The goal of this RFC is to define an external-facing API for Linkerd that:
+1. Cluster operators need guidance on how to support Linkerd's Prometheus
+   data. There should be guidance on how to manage cardinality (especially in
+   larger clusters) and how to support multi-tenant environemnts.
+2. Consumers have inadequate documentation and tools to access Linkerd's
+   metrics data. The proxy's metrics export has been treated as an
+   implementation detail and not a public API. Information is exposed in such a
+   way that a consumer essentially has to understand how the proxy is
+   implemented to know how to interpret the data.
 
-1. The semantics of which are defined by this RFC;
-2. Is stable/versioned;
-3. Supports Linkerd's UIs as well as arbitrary external clients;
-4. Supports Kubernetes RBAC;
-5. Supports extensions for additional protocols, metrics, and metadata; and
-6. Supports arbitrary workload hierarchies (i.e for custom operators).
-
-## Design Notes
-
-### API
-
-The metrics API is read-only.
-
-In order to support RBAC, all requests should be scoped to the resources on which metrics are
-measured.
-
-It is time-oriented. It should have both time-series and summary capabilities.
-
-Each Histogram should be given sufficient, published defaults.
-
-#### TCP
-
-- Open Count, by
-  - Interface: `inbound | outbound`
-  - Workload
-  - Service
-  - Service profile route
-  - Peer Identity
-  - Peer K8s Labels? How?
-- Close Count, by the above as well as an errno.
-- Connect Latency Histogram, by the above
-- Connection Age Histogram, by the above
-- Bytes Transferred, by the above
-
-#### HTTP
-
-- Request count, by
-  - Interface: `inbound | outbound`
-  - Authority
-  - Workload
-  - Service
-  - Service profile route
-  - Peer Identity
-  - Peer K8s Labels? How?
-- Response count, by
-  - The above, plus
-  - Status Code; [100, 599]
-  - Classification: success | failure
-- Response Latency Histogram
-  - Should users be able to query at arbitrary percentiles?
-
-##### gRPC
-
-How do we represent the gRPC-specific metadata that Linkerd exposes? It seems like it should be
-distinct from (but in addition to) the HTTP-level abstractions? Or should HTTP metrics have a
-hole to punch through arbitrary metadata?
-
-Should we provide gRPC stats scoped by gRPC service name? We can infer these via known path
-structure when the appropriate headers are present.
-
-#### K8s services
-
-One of the main complexities in our metrics comes from handling Service-specific metadata.
-Currently all service-specific metadata is measured on the client-side, even when the querier
-asks about server-side resources. It is effectively impossible to query for server-side metrics
-for a service, since we allow pods to exist in more than one service. This is
-[fixable][rfc-src-metrics], however, if servers are able to discover metadata for inbound
-traffic.
-
-It seems like we should not support querying service metrics as a resource scope, though, because
-we cannot measure traffic _on_ a Service. We can measure outbound traffic from a set of client
-pods to a Service; and we can measure inbound traffic to a set of server pods that is targetted
-at a Service. I think our API should be explicit about these constraints.
-
-[rfc-src-metrics]: https://github.com/linkerd/rfc/pull/15
-
-### Prior Art
-
-#### Linkerd Stat
-
-Linkerd's [`Stat`][stat] API is resource oriented, but it only exposes an _outbound_ side, so
-there's no way to
-
-Why is `skip_stats` part of the API?
-
-The `tcp_stats` flag is a legacy of the CLI-specific nature of the command... it seems like we
-should either support a list of protocols or have protocol-specific endpoints.
-
-[l5d-stat]: https://github.com/linkerd/linkerd2/blob/dacf87e084a55757aab9cfa8556053495e43e207/proto/public.proto#L348-L427
-
-#### SMI
-
-[SMI's metrics][smi-spec] appear to be stringly typed, in that their are a few well-defined
-(http-specific?) metrics, but the base set of metrics is insufficient to satisfy Linkerd's CLI
-needs.
-
-However, the way it's resource-oriented is very appealing.
-
-It does not yet expose any time-series functionality.
-
-[smi-spec]: https://github.com/servicemeshinterface/smi-spec/blob/d5dd526a1a19c784065dc3fbac05a52decc01735/traffic-metrics.md
-
-#### Prometheus
-
-Can we expose a [Prometheus API][prom-api] scoped under each resource? E.g.
-`.../v1/deploy/deployname-xyzxb/query?...` and `.../v1/deploy/deployname-xyzxb/metrics`?
-
-We would conceivably parse, validate/authorize, and translate prometheus queries to be issued to
-the backing store. We would define a well-defined set of public metrics/labels.
-
-This could hit the sweet spot of interopability with scoping for access control.
-
-[prom-api]: https://prometheus.io/docs/prometheus/latest/querying/api/
+The goal of this RFC is to define an external-facing metrics API for Linkerd.
+This API should be stable, versioned, and extensible to support additional
+protocols, metrics, and metadata (via future RFCs).
--- a/problem/0000-metrics-v1.md
+++ b/problem/0000-metrics-v1.md
@@ -1,0 +1,136 @@
+# Linkerd Metrics v1
+
+- Contribution Name: metrics-v1
+- Implementation Owner:
+- Start Date: 2020-04-23
+- Target Date:
+- RFC PR:
+- Linkerd Issue:
+- Reviewers: @grampelberg @adleong
+
+## Problem Statement
+
+[problem-statement]: #problem-statement
+
+Linkerd's `stat` API powers both the command-line and web interfaces, but it's tailored closely
+to their use cases and it has evolved only to express the features of these tools. Because of its
+narrow focus, integrators tend to read data directly from Prometheus, which features an
+expressive query language and a variety of external integrations/extensions. This presents an
+implicit, undocumented relationship between the proxy's metrics exposition structure and traffic
+in Kubernetes.
+
+The goal of this RFC is to define an external-facing API for Linkerd that:
+
+1. The semantics of which are defined by this RFC;
+2. Is stable/versioned;
+3. Supports Linkerd's UIs as well as arbitrary external clients;
+4. Supports Kubernetes RBAC;
+5. Supports extensions for additional protocols, metrics, and metadata; and
+6. Supports arbitrary workload hierarchies (i.e for custom operators).
+
+## Design Notes
+
+### API
+
+The metrics API is read-only.
+
+In order to support RBAC, all requests should be scoped to the resources on which metrics are
+measured.
+
+It is time-oriented. It should have both time-series and summary capabilities.
+
+Each Histogram should be given sufficient, published defaults.
+
+#### TCP
+
+- Open Count, by
+  - Interface: `inbound | outbound`
+  - Workload
+  - Service
+  - Service profile route
+  - Peer Identity
+  - Peer K8s Labels? How?
+- Close Count, by the above as well as an errno.
+- Connect Latency Histogram, by the above
+- Connection Age Histogram, by the above
+- Bytes Transferred, by the above
+
+#### HTTP
+
+- Request count, by
+  - Interface: `inbound | outbound`
+  - Authority
+  - Workload
+  - Service
+  - Service profile route
+  - Peer Identity
+  - Peer K8s Labels? How?
+- Response count, by
+  - The above, plus
+  - Status Code; [100, 599]
+  - Classification: success | failure
+- Response Latency Histogram
+  - Should users be able to query at arbitrary percentiles?
+
+##### gRPC
+
+How do we represent the gRPC-specific metadata that Linkerd exposes? It seems like it should be
+distinct from (but in addition to) the HTTP-level abstractions? Or should HTTP metrics have a
+hole to punch through arbitrary metadata?
+
+Should we provide gRPC stats scoped by gRPC service name? We can infer these via known path
+structure when the appropriate headers are present.
+
+#### K8s services
+
+One of the main complexities in our metrics comes from handling Service-specific metadata.
+Currently all service-specific metadata is measured on the client-side, even when the querier
+asks about server-side resources. It is effectively impossible to query for server-side metrics
+for a service, since we allow pods to exist in more than one service. This is
+[fixable][rfc-src-metrics], however, if servers are able to discover metadata for inbound
+traffic.
+
+It seems like we should not support querying service metrics as a resource scope, though, because
+we cannot measure traffic _on_ a Service. We can measure outbound traffic from a set of client
+pods to a Service; and we can measure inbound traffic to a set of server pods that is targetted
+at a Service. I think our API should be explicit about these constraints.
+
+[rfc-src-metrics]: https://github.com/linkerd/rfc/pull/15
+
+### Prior Art
+
+#### Linkerd Stat
+
+Linkerd's [`Stat`][stat] API is resource oriented, but it only exposes an _outbound_ side, so
+there's no way to
+
+Why is `skip_stats` part of the API?
+
+The `tcp_stats` flag is a legacy of the CLI-specific nature of the command... it seems like we
+should either support a list of protocols or have protocol-specific endpoints.
+
+[l5d-stat]: https://github.com/linkerd/linkerd2/blob/dacf87e084a55757aab9cfa8556053495e43e207/proto/public.proto#L348-L427
+
+#### SMI
+
+[SMI's metrics][smi-spec] appear to be stringly typed, in that their are a few well-defined
+(http-specific?) metrics, but the base set of metrics is insufficient to satisfy Linkerd's CLI
+needs.
+
+However, the way it's resource-oriented is very appealing.
+
+It does not yet expose any time-series functionality.
+
+[smi-spec]: https://github.com/servicemeshinterface/smi-spec/blob/d5dd526a1a19c784065dc3fbac05a52decc01735/traffic-metrics.md
+
+#### Prometheus
+
+Can we expose a [Prometheus API][prom-api] scoped under each resource? E.g.
+`.../v1/deploy/deployname-xyzxb/query?...` and `.../v1/deploy/deployname-xyzxb/metrics`?
+
+We would conceivably parse, validate/authorize, and translate prometheus queries to be issued to
+the backing store. We would define a well-defined set of public metrics/labels.
+
+This could hit the sweet spot of interopability with scoping for access control.
+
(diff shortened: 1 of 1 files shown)


problem/0000-metrics-v1.md
```diff
@@ -0,0 +1,136 @@
...
+[problem-statement]: #problem-statement
+
+Linkerd's `stat` API powers both the command-line and web interfaces, but it's tailored closely
+to their use cases and it has evolved only to express the features of these tools. Because of its
+narrow focus, integrators tend to read data directly from Prometheus, which features an
+expressive query language and a variety of external integrations/extensions. This presents an
+implicit, undocumented relationship between the proxy's metrics exposition structure and traffic
+in Kubernetes.
+
+The goal of this RFC is to define an external-facing API for Linkerd that:
```
adleong: It would be helpful to be explicit about where our current metrics API falls short of these requirements.

adleong: 1. The current API has well defined semantics (though I'm sure the documentation of those semantics could be improved)
2. As far as I know, we haven't published any guarantees or guidelines about the stability of our API, but we also have not yet made any backwards incompatible changes.
3. What does it mean to support external clients?  Is there specific functionality that needs to be satisfied?
4. The current API does not check RBAC, though I believe the structure of the API is well suited to adding this.
5. Can you explain what this means?
6. Can you explain what this means?

problem/0000-metrics-v1.md
```diff
@@ -0,0 +1,136 @@
...
+5. Supports extensions for additional protocols, metrics, and metadata; and
+6. Supports arbitrary workload hierarchies (i.e for custom operators).
+
+## Design Notes
+
+### API
+
+The metrics API is read-only.
+
+In order to support RBAC, all requests should be scoped to the resources on which metrics are
```
adleong: I like this mapping of RBAC onto the point at which the metrics are measured :+1:

problem/0000-metrics-v1.md
```diff
@@ -0,0 +1,136 @@
...
+measured.
+
+It is time-oriented. It should have both time-series and summary capabilities.
```
adleong: Are we sure this is a requirement?  I can see this being necessary if API consumers are trying to build charting or historical lookups on top of this API, but is this something we want to support?

problem/0000-metrics-v1.md
```diff
@@ -0,0 +1,136 @@
...
+
+Each Histogram should be given sufficient, published defaults.
```
adleong: I don't understand what this means.

problem/0000-metrics-v1.md
adleong: How much history is available from this API?  How can users extend the amount of history available?

Today, users can federate Linkerd's Prometheus into their own Prometheus with a longer retention period.  Is it a goal of this RFC that users would no longer federate directly from Linkerd's Prometheus?  Does this API support the use-case of federating Linkerd's metrics into longer term storage?

problem/0000-metrics-v1.md
```diff
@@ -0,0 +1,136 @@
...
+  - Service
+  - Service profile route
+  - Peer Identity
+  - Peer K8s Labels? How?
+- Response count, by
+  - The above, plus
+  - Status Code; [100, 599]
+  - Classification: success | failure
+- Response Latency Histogram
+  - Should users be able to query at arbitrary percentiles?
```
adleong: As long as we provide reasonable percentiles, I can't see a compelling reason why users would want this.

problem/0000-metrics-v1.md
```diff
@@ -0,0 +1,136 @@
...
+How do we represent the gRPC-specific metadata that Linkerd exposes? It seems like it should be
+distinct from (but in addition to) the HTTP-level abstractions? Or should HTTP metrics have a
+hole to punch through arbitrary metadata?
+
+Should we provide gRPC stats scoped by gRPC service name? We can infer these via known path
+structure when the appropriate headers are present.
+
+#### K8s services
+
+One of the main complexities in our metrics comes from handling Service-specific metadata.
```
adleong: By service-specific metadata, do you mean traffic metrics for a service resource?  Or metrics which are measured at the logical (service) level?  Both of these can be measured and queried on the server-side by today's API.  

The API request that corresponds to the `linkerd routes deploy/foo` command requests server-side logical (route) metrics from pods in the foo deployment.
The API request that corresponds to the `linkerd stat svc/foo` command requests server-side metrics for requests addressed to the `foo` service (according to their authorities).

problem/0000-metrics-v1.md
```diff
@@ -0,0 +1,136 @@
...
+Currently all service-specific metadata is measured on the client-side, even when the querier
+asks about server-side resources. It is effectively impossible to query for server-side metrics
+for a service, since we allow pods to exist in more than one service. This is
+[fixable][rfc-src-metrics], however, if servers are able to discover metadata for inbound
```
adleong: I don't think this requires the linked RFC.  On the server-side we record the request's authority, so we know which service is was addressed to.

problem/0000-metrics-v1.md
```diff
@@ -0,0 +1,136 @@
...
+traffic.
+
+It seems like we should not support querying service metrics as a resource scope, though, because
```
adleong: I agree with this and it meshes nicely with the invariant that RBAC should be tied to the resource at which the metrics were recorded.

problem/0000-metrics-v1.md
```diff
@@ -0,0 +1,136 @@
...
+at a Service. I think our API should be explicit about these constraints.
+
+[rfc-src-metrics]: https://github.com/linkerd/rfc/pull/15
+
+### Prior Art
+
+#### Linkerd Stat
+
+Linkerd's [`Stat`][stat] API is resource oriented, but it only exposes an _outbound_ side, so
+there's no way to
```
adleong: the stat API exposes inbound metrics when neither `to_resource` not `from_resource` are specified.

problem/0000-metrics-v1.md
```diff
@@ -0,0 +1,136 @@
...
+
+Why is `skip_stats` part of the API?
```
adleong: I'm not sure.  This seems to be used extensively in the Linkerd dashboard, but I don't understand the dashboard code well enough to figure out why.

problem/0000-metrics-v1.md
```diff
@@ -0,0 +1,136 @@
...
+
+However, the way it's resource-oriented is very appealing.
+
+It does not yet expose any time-series functionality.
+
+[smi-spec]: https://github.com/servicemeshinterface/smi-spec/blob/d5dd526a1a19c784065dc3fbac05a52decc01735/traffic-metrics.md
+
+#### Prometheus
+
+Can we expose a [Prometheus API][prom-api] scoped under each resource? E.g.
```
adleong: This is very appealing and plays very nicely with RBAC.

problem/0000-metrics-v1.md
adleong: Is it a goal of this RFC to add access control to Prometheus?  Should accessing Prometheus directly be an available but unsupported/discouraged API?  Or should it be locked down in some way?

problem/0000-metrics-v1.md
olix0r: The proxy's histogram windows were chosen ~arbitrarily and probably hide meaningful data. This decision needs to be revisited as we publish the API.

problem/0000-metrics-v1.md
olix0r: At an API level there are no restrictions. Operators are able to tune/configure this to their needs. The API returns whatever data is available in the underlying storage.  We will need to publish docs/config/tools/etc for Linkerd scrape configs that provided the metrics needed to power the API (though this is out of scope for this RFC, imo... this should focus just at the API-level)

problem/0000-metrics-v1.md
olix0r: Why is this a necessary or helpful limitation though?

problem/0000-metrics-v1.md
olix0r: I'm talking about metrics for concrete services. These can only be measured on the clientside today.

problem/0000-metrics-v1.md
olix0r: Authority indicates logical service, but we lose concrete services.

problem/0000-metrics-v1.md
olix0r: I find this behavior very surprising and difficult to reason about. I'm looking for ways to make the api much more explicit.  The fact that we have an `outbound` query makes it seem like there should have `inbound` query options that limit inbound stats (but we don't have the labels here to satisfy these requests yet).

problem/0000-metrics-v1.md
olix0r: I think we want to provide guidance/tools that allows folks to lock down their prometheus. I think we should consider locking down our own prometheus as we grow those capabilities. But I don't think it's desirable that we implement access control by default. Not yet, anyway. 

Really, what I'm trying to say in this part is: as we look at our API options, let's ensure that we can satisfy at least these needs. The impl plan will, of course, order this incrementally.

problem/0000-metrics-v1.md
olix0r: Specifically, percentiles mean very different things at different traffic levels. Linkerd exposes raw histograms, not percentile summaries. It seems like a strength that we should expose and not hide.

problem/0000-metrics-v1.md
olix0r: I think so. Even though linkerd's interfaces only expose summary data, it seems extremely desirable to expose timeseries. Otherwise, we're basically forcing all of these needs through the raw underlying prometheus interface. Though, from the impl-plan point of view I think we could avoid doing this right away, I think we should plan for the API to satisfy these needs.

problem/0000-metrics-v1.md
olix0r: The current APIs are undocumented. The ambiguities you discuss in the source-metrics RFC are exactly what I mean by the semantics not being well-defined. We have no way to reason about what exactly is being measured/reported but for reading/understanding the implementation. One potential outcome for this RFC is that we simply decide that we like the `Stat` API, document it, and provide examples of how to integrate with it. However, I don't think that belongs in the problem statement. I want to approach the metrics API with fresh eyes and to document exactly how it should work so that we have a plan for how it will grow. So that when we want to extend it, there's a clear story for doing so.

But the root problem we're solving is, as I mentioned, that it's far easier for users to just go to prometheus and get data, and we have absolute no documentation/guarantees/etc about this implicit API.

> What does it mean to support external clients? Is there specific functionality that needs to be satisfied?

This is more of a product requirement than a technical one. I think gRPC may be a high bar for integrators (though it's certainly an option on the table). I also think we should be looking at our options in terms of interop. If, for instance, we can expose a version of prometheus's API, we could support many external clients with very little additional work.

> 5. Supports extensions for additional protocols, metrics, and metadata; and

What happens if we add Kafka support, where does that go?

What happens when we expose new proxy metrics (like http errors, by reason), where should that go?

Basically, we need an extensibility story.

> 6. Supports arbitrary workload hierarchies (i.e for custom operators).

You can think of a `Deployment` as a CRD for the deployment operator. It knows how to create ReplicaSets, which triggers it to create Pods (effectively, anyway).

People are creating other, bespoke workload types that create pods that are owned by alternate, custom resource types. We should have a story for supporting these kinds of Kubernetes extensions.

problem/0000-metrics-v1.md
olix0r: And, it's likely that different histograms should have different defaults. I.e. connection_duration needs very different resolution than response_latency

problem/0000-metrics-v1.md
```diff
@@ -0,0 +1,136 @@
...
+
+It is time-oriented. It should have both time-series and summary capabilities.
+
+Each Histogram should be given sufficient, published defaults.
+
+#### TCP
+
+- Open Count, by
+  - Interface: `inbound | outbound`
+  - Workload
```
olix0r: I wonder if it makes sense to expose Node information as well? To help correlate node issues

problem/0000-metrics-v1.md
```diff
@@ -0,0 +1,136 @@
...
+
+The goal of this RFC is to define an external-facing API for Linkerd that:
+
+1. The semantics of which are defined by this RFC;
+2. Is stable/versioned;
+3. Supports Linkerd's UIs as well as arbitrary external clients;
+4. Supports Kubernetes RBAC;
+5. Supports extensions for additional protocols, metrics, and metadata; and
+6. Supports arbitrary workload hierarchies (i.e for custom operators).
+
```
grampelberg: Couple questions around requirements:

- Should we add a requirement around the query interface? I've been thinking about what kinds of queries users will want to run (the linkerd UI is a great place to start there).
- Should there be a requirement around published clients? Having an API is great, publishing a client in your language of choice is way easier to use and consume.
- What kind of assumptions are we going to make about the backend data store? Should it be pluggable?

problem/0000-metrics-v1.md
grampelberg: Do we want to be explicit about the RBAC policies? It could really run the gamut. I suspect that almost anything we do can be improved in the future, so it is more of an exercise around "out of scope" at this point.

problem/0000-metrics-v1.md
grampelberg: I was thinking about this a little bit too. Ideally, we'd not bother and it would always be point in time. Push it back to the implementation to do time series.

Unfortunately, having this data be point in time complicates the usage. Imagine you're scraping the API and putting it in your own time-series database. Scraping on odd intervals means that you'll get imperfect data and have to reason about quite a bit in the eventual database. If we are okay requiring the backends for this API to be time-series aware, it seems like a good UX improvement to deliver some kind of time series data back to the end user.

problem/0000-metrics-v1.md
grampelberg: In a perfect world, we'd get labels for everything recursively from the pod up to the final parent.

problem/0000-metrics-v1.md
```diff
@@ -0,0 +1,136 @@
...
+Each Histogram should be given sufficient, published defaults.
+
+#### TCP
+
+- Open Count, by
+  - Interface: `inbound | outbound`
+  - Workload
+  - Service
+  - Service profile route
```
grampelberg: How would we have routes here?

problem/0000-metrics-v1.md
```diff
@@ -0,0 +1,136 @@
...
+  - Peer Identity
+  - Peer K8s Labels? How?
+- Close Count, by the above as well as an errno.
+- Connect Latency Histogram, by the above
+- Connection Age Histogram, by the above
+- Bytes Transferred, by the above
```
grampelberg: Any thoughts on how this'll degrade? I'd love to have a query for ingress/egress but that wouldn't have in-mesh peers.

problem/0000-metrics-v1.md
```diff
@@ -0,0 +1,136 @@
...
+- Response count, by
+  - The above, plus
+  - Status Code; [100, 599]
+  - Classification: success | failure
+- Response Latency Histogram
+  - Should users be able to query at arbitrary percentiles?
+
+##### gRPC
+
+How do we represent the gRPC-specific metadata that Linkerd exposes? It seems like it should be
```
grampelberg: This is a great opportunity to flesh out what metrics for another protocol would look like.

problem/0000-metrics-v1.md
```diff
@@ -0,0 +1,136 @@
...
+distinct from (but in addition to) the HTTP-level abstractions? Or should HTTP metrics have a
+hole to punch through arbitrary metadata?
+
+Should we provide gRPC stats scoped by gRPC service name? We can infer these via known path
```
grampelberg: Do we have any cardinality concerns with gRPC? Would it be possible to just collect per-method metrics by default?

problem/0000-metrics-v1.md
```diff
@@ -0,0 +1,136 @@
...
+- Close Count, by the above as well as an errno.
+- Connect Latency Histogram, by the above
+- Connection Age Histogram, by the above
+- Bytes Transferred, by the above
+
+#### HTTP
+
+- Request count, by
+  - Interface: `inbound | outbound`
+  - Authority
```
grampelberg: Do we want to differentiate between the *real* authority and the normalized one? Feels like it'd be valuable to lookup `foo.com` as an authority even though it ended up at the normalized `foo.default.svc.cluster.local`.

problem/0000-metrics-v1.md
```diff
@@ -0,0 +1,136 @@
...
+Why is `skip_stats` part of the API?
+
+The `tcp_stats` flag is a legacy of the CLI-specific nature of the command... it seems like we
+should either support a list of protocols or have protocol-specific endpoints.
+
+[l5d-stat]: https://github.com/linkerd/linkerd2/blob/dacf87e084a55757aab9cfa8556053495e43e207/proto/public.proto#L348-L427
+
+#### SMI
+
+[SMI's metrics][smi-spec] appear to be stringly typed, in that their are a few well-defined
```
grampelberg: `strongly`?

problem/0000-metrics-v1.md
grampelberg: I'm wondering if we can get everything we want from recording rules. That provides the stable API and solves some other issues. We'd still need to parse and such, but the surface area would be shrunk.

problem/0000-metrics-v1.md
adleong: For what it's worth, when you have a chain of federated Prometheuses, each Prometheus exposes point-in-time data to the next Prometheus in the chain which scrapes it at regular intervals.  So it's not unheard of.

I'm not sure if I'm reasoning about this correctly, but I think if you have mismatched scrape intervals, the resolution of the data will effectively be the longest scrape interval.  I don't think the intervals will compound in some wonky way.

That said, it seems to me like there are 3 different formats our API could expose metrics in: 
* summaries over a time period (as the gRPC API does today)
* point-in-time snapshot of gauges/counters/histograms (ala hitting the /metrics or /federate endpoint)
* timeseries (like what is returned from a promQL query)

We need to decide which of these we want to support.

adleong: Users today who want a retention period of longer than 6 hours federate the Linkerd prometheus into their own.  If the plan is to declare that Linkerd's Prometheus is not part of its API then I think we need to figure out how to support this use-case.  

e.g. Should the API support bulk extraction of metrics into an external longer-term storage?  Or should Linkerd API surface area include knobs for increasing the internal retention period?

problem/0000-metrics-v1.md
adleong: No arguments here :)

problem/0000-metrics-v1.md
olix0r: @adleong That all makes sense to me.

I think that if we can support the generic time-series case, the summary case becomes ~trivial (it's a `sum` or `rate` query over the interval, right?).

problem/0000-metrics-v1.md
adleong: The more I think about this, the more I like it.  It also makes the transition easier for clients who are currently scraping Prometheus directly.

problem/0000-metrics-v1.md
adleong: Yeah, assuming the API supports this (as promQL does).

problem/0000-metrics-v1.md
olix0r: Recording rules are probably a good way to try this, yeah.

But, I think there are opportunities for us to be a bit clever about our storage/query patterns to take advantage of our understanding of the data model.

For example: Can we reduce the number of time-series that have to be stored/in memory by _only_ storing the pod name in a metric label? What if, instead, we wrote a single `pod{name="",parent_kind="...",parent_name="...",label_foo="...",label_bar="...} 1`. If you want to do a deploy-specific lookup, instead of looking up a `request_total{deployment="foo"}`, we would first expand the list of pods for the deployment and look up by `{pod=".."}`...

This is, of course, totally hypothetical and we'd want to dig on access patterns before optimizing this way. But my point is this plan gives us enough room to get creative...

problem/0000-metrics-v1.md
olix0r: I feel like this is generally small cardinality, but we could always put explicit limits in...

problem/0000-metrics-v1.md
olix0r: I don't have a good mental model for thinking about this yet. We need a clearer idea of what resources express these mappings...

problem/0000-metrics-v1.md
olix0r: I'm mostly trying to keep this at doc at the interface-level, but I will dig in more on query interface.

problem/0000-metrics-v1.md
olix0r: Yeah, the design should provide an initial RBAC model. I don't think we should get into going so granular as to which metrics can be read... You can either read metrics for a pod or you can't.

problem/0000-metrics-v1.md
olix0r: I need to think some more about routes & retries... I'd like to eliminate some redundancy in how we export these metrics.

problem/0000-metrics-v1.md
olix0r: I'm cagey about exposing per-source-IP information for Internet-facing servers. We should probably call out ingress/egress explicitly. We want to optimize for in-mesh visibility, of course, but there's a balance worth striking there. At the very least we want to differentiate TLS/SNI vs plaintext.

problem/0000-metrics-v1.md
olix0r: I've got a nagging suspicion that this isn't going to work out all that well. Or that it will be wildly complex.

What happens if we have something like `../ns/linkerd/query`? How does RBAC apply? Must I have permission to read metrics for each pod/resource in the namespace? Or do I require an explicit RBAC binding on the namespace itself? Am I able to write a query that spans namespaces? How would that work?

What if we don't try to enforce RBAC on prometheus's _data_ at all?

A preferable model might be:

* Prometheus scrapes themselves are RBAC'd. I.e., the proxy applies access control to its own /metrics endpoints.
* The prometheus instance is the unit of view isolation. A user who can read from a prometheus instance has permission to read all of its data. If multi-tenancy is needed, users should deploy a Prometheus instance fore each tenant. RBAC is used to grant each instance permission to read metrics from the appropriate resources.
* Linkerd's challenge is then:
  * RBAC enforcement (for /metrics & prometheus access)
  * When a user interacts with Linkerd UIs how do we use the prometheus instance the user has access to?
    * Perhaps we make it easy to deploy a subset of the control plane per tenant? There's no good reason the `web` instance needs to be centralized in this world.
  * We publish a prometheus "schema" and scrape configs...

Basically: "Bring your own prometheus" should support arbitrary numbers of prometheus instances. Our users who really truly want to dump all of their data into one huge prometheus instance don't get view isolation (except by limiting who can read data from their monoprom).

olix0r:
@adleong @grampelberg I dropped all of the design notes and just left this with a simpler problem statement. Please ACK if you're good with this. I'll proceed with the meat of the proposal.

grampelberg (approved):
LGTM

olix0r:

PR #25, opened 2020-05-12
Initial pass at protocol detection

grampelberg:


.prettierrc
```diff
@@ -0,0 +1 @@
+proseWrap: always
```
hawkw, line 1: was this supposed to be committed?

problem/0005-protocol-detection.md
```diff
@@ -0,0 +1,189 @@
...
+
+Users should be able to get the benefit of lower level protocols (L4) when
+Linkerd is unable to detect the higher level protocol (L7). This allows users to
+get all the benefits they would expect from mTLS, TCP metrics and multi-cluster
+routing. To do this, users need to specify whether protocol detection should be
+skipped on a per-port basis.
+
+### Requirements
+
+- Allow ports to skip the proxy entirely (inbound and outbound), maintaining
```
adleong, line 44: In what situations should the proxy be skipped entirely vs just skipping protocol detection?  Protocol detection should be skipped for known SSF protocols and proxy should be skipped when the proxy is suspected of causing problems (i.e. solely for diagnosing and working around bugs)?

problem/0005-protocol-detection.md
```diff
@@ -0,0 +1,189 @@
...
+```
+
+If appProtocol is specified and not on the list of supported protocols, the
+proxy would skip protocol detection.
+
+#### Downsides
+
+- This is currently an alpha feature (in 1.18). It is basically unusable at this
+  point in time.
+- Requires the destination to do discovery on what services are pointed at it
```
adleong, line 87: The destination already does discovery for the purposes of service profiles.  @olix0r is inbound discovery synchronous?  that is, is the first inbound request blocked until discovery is complete?

adleong (approved):
This looks great to me.  I have a preference for option 2 or 3 since it doesn't limit our kubernetes version compatibility.

problem/0005-protocol-detection.md
```diff
@@ -0,0 +1,189 @@
...
+    app: mysql
+  clusterIP: None
+```
+
+Similar to `config.linkerd.io/skip-outbound-ports`, the
+`config.linkerd.io/skipDetection` annotation would take either a range or comma
+separated list of ports to skip detection on for this service.
+
+#### Downsides
+
```
adleong, line 117: I think this also misses traffic which is sent directly to a pod IP without using a service.

problem/0005-protocol-detection.md
olix0r: We know that there are scenarios where users prefer to skip, for instance, 443, since the proxy adds extremely little value for these streams; especially when startup-ordering issues remain a reality.

Administrators can always disable this feature by adding gatekeeper-style policy enforcements in place to prevent such annotations from being set.

Are there specific concerns?

problem/0005-protocol-detection.md
olix0r: Yes, though this would significantly expand this to cover all TCP traffic, etc.

olix0r: Also, we'd need more concretely about what it means to couple this with service profiles... it would be a substantial change to how the proxy does discovery.

problem/0005-protocol-detection.md
```diff
@@ -0,0 +1,189 @@
...
+- Allow ports to skip only protocol detection inside the proxy (inbound and
+  outbound).
+- Support port ranges as well as individual ports.
+- Surface metrics at the TCP level.
+- Work for applications where both the client and server are meshed.
+- Do not require configuration on both the client and server.
+
+### Out of Scope
+
+- Configuration of the protocol on a port.
```
olix0r, line 55: What does this mean?

olix0r (approved):
Big +1 on problem. I have some thoughts on a proposal...

problem/0005-protocol-detection.md
adleong: I don't have specific concerns.  These two features have some amount of overlap so I just want to make sure we have a good handle on which situations to use each one.

problem/0005-protocol-detection.md
```diff
@@ -0,0 +1,189 @@
...
+```
+
+Similar to `config.linkerd.io/skip-outbound-ports`, the
+`config.linkerd.io/skipDetection` annotation would take either a range or comma
+separated list of ports to skip detection on for this service.
+
+#### Downsides
+
+- Annotations are unstructured, not in-line and specific to Linkerd.
+- Same discovery issues as option 1.
```
adleong, line 119: How would this work for inbound TCP traffic?  would we do discovery on the SO_ORIG_DST?  would the SO_ORIG_DST be the service cluster IP or pod IP in this case?

olix0r:
I'll write this up in a PR, but here's my general proposal:

* We opt for workload annotations. It's simplest (there's no real need for this to be dynamically configurable on the inbound side -- it really is a matter of annotating how to configure the workload).
* Annotations in the form of something like `protocols.linkerd.io/[protocol]: [port-list]`, e.g. `protocols.linkerd.io/tcp: 4444,5555`, `protocols.linkerd.io/http1: 8000-8999`
* Supported protocols (at first) include `tcp`, `tls`, `http1`, `h2`
  * `tcp`: no detection is performed and stream is forwarded unmodified (after any mtls concerns, out of scope for now).
  * `tls`: the application performs its own TLS. we can safely analyze TLS messages for e.g. SNI data.
  * `http1`: http/1.0 or http/1.1 -- we give connections directly to hyper without detection
  * `h2`: http/2 (and grpc) -- we give connections directly to hyper without detection
* For Inbound
  - workload configured at startup via injector/env
  - accept stack modified to use configuration-based protocol detection.
  - if no protocol specified, existing detection logic is used.
* For outbound
  - Accept stack modified to do discovery _in parallel with_ detection (cached by target ip:port)
  - We'll probably want to use the normal Destinations.Get endpoint; and we probably want to figure out how to build the HTTP LB's off of this discovery stream without having to go back to the network. This could also be an opportune time to think about how this interacts with admin endpoints that dump this state. Needs to be RAII-managed, etc, to prevent leaks.
  - The real wrinkle here is that we'll need to make this decision for a _service_, but all of the configuration is on the _endpoints_ in that service. Basically, it's a bug if all of the endpoints in a service do not have the same protocol. We could get fancy about enforcing this at inject-time, but we probably need to be careful about how we handle this in the destination service. It would be ideal if inconsistencies were handled as gracefully as possible (i.e. if set on some subset of endpoints and unset otherwise, assume setting for all). Needs some thought on diagnostics...

I think the inbound path for this is fairly straightforward. The outbound path is more complex, and is somewhat coupled with the tcp lb for mtls.

.prettierrc
grampelberg: Yeah, I guess I could split it out into another PR but it sure makes life easier.

problem/0005-protocol-detection.md
grampelberg: I think it is pretty important to have a "break glass in case of emergency" feature. This'll make it so we can downplay port skipping itself but I think it'll be a valuable tool for getting folks out of an incident.

problem/0005-protocol-detection.md
grampelberg: That's a great point! I'll add it in.

problem/0005-protocol-detection.md
grampelberg: Calling a port `http`, `grpc` or `kafka`. I originally called this in scope and realized that I didn't have good enough of a use case in mind to make it all fit together.

admc:
👍

PR #30, opened 2020-06-14
Proposal for circuit breaking functionality

Ashish-Bansal:
RFC for adding circuit breaking functionality -- https://github.com/linkerd/linkerd2/issues/2846.

Initial thoughts for circuit breaking based on limited context I have about linkerd2's architecture. 

[Rendered](https://github.com/Ashish-Bansal/rfc/blob/circuit-breaking/design/0005-circuit-breaking.md)

### DIFF
--- a/design/0005-circuit-breaking.md
+++ b/design/0005-circuit-breaking.md
@@ -1,0 +1,150 @@
+# Title
+
+- Contribution Name: circuit-breaking
+- Implementation Owner: @Ashish-Bansal, Linkerd Team
+- Start Date: 2020-06-14
+- Target Date: N.A.
+- Linkerd Issue:
+  [linkerd/linkerd2#2846](https://github.com/linkerd/linkerd2/issues/2846)
+- Reviewers: @adleong
+
+## Summary
+
+[summary]: #summary
+
+We would like to implement circuit breaking functionality which would allow end
+users to configure mesh such that they can limit the impact of failures from
+faulty service over others and prevent cascading issues. By configuring circuit
+breaking, they will be able to remove failing service instances from load
+balancing and survive through it in a more graceful manner.
+
+## Problem Statement
+
+In a service mesh, services often make calls to other services. In case upstream
+service is unable to respond due to any reason, it's possible that these
+failures can cascade to other services in the cluster.
+
+If we are aware that specific service instance is not responding, we can call it
+unhealthy and it's better if we just stop sending "traffic"[0] to that instance
+rather than getting just another expected failure.
+
+Users should also be able to set max-ejection-percentage which provides max
+number of service instances which can be removed from load balancing for
+specific service.
+
+Along with that, if we detect that "almost all" the instances of service are not
+responding, then users should be able to cut off traffic to whole service as
+well. To quantize "almost all" term, user should be able to configure how many
+percentage of service instances would be removed from the load balancing list
+before whole traffic to that service would be circuit-breaked. This is in
+constrast with max-ejection-percentage and would be mutually exclusive.
+
+User should also be able to select ejection policy based on which any instance
+would be ejected from load balancing.
+
+It would also be somewhat helpful in case of canary deployments, when there's
+some regression and new build is raising failures.
+
+[0] The term 'traffic' means any kind of communication happening between two
+services. It can be HTTP requests or gRPC messages over HTTP or L4 TCP
+connections. But the circuit breaking functionality itself won't be protocol
+agnostic. Its implementation would be protocol specific and it's upto community
+which protocol they want to add support for and how circuit breaking will make
+sense in that protocol.
+
+## Design proposal
+
+[design-proposal]: #design-proposal
+
+Right now if I just consider ServiceProfile spec, IMO it seems to represent more
+of profile-for-http-service rather than profile-for-generic-service. May be it
+makes sense given that it supports only HTTP and grpc over HTTP as of now. Given
+current architecture, I feel it's best to provide circuit breaking functionality
+through service profiles. That means it would be limited to above two kind of
+services as of now.
+
+For TCP connections, according to original github
+[issue](https://github.com/linkerd/linkerd2/issues/2846#issue-447856936), it
+seems like linkerd2 does support automatic circuit breaking for TCP errors but
+I'm not sure of its behaviour -- whether TCP error means connection refused or
+resets or timeouts or something else. Anyway, it seems like it's not
+configurable and I'm not sure if making it configurable has any important
+usecases.
+
+In future, for any other protocol support for circuit breaking, one will first
+need to incorporate the protocol into ServiceProfile itself and then implement
+the circuit breaking functionality.
+
+Regarding ejection of the service instance, once we decide any instance needs to
+be ejected in accordance with mentioned configuration, it would be ejected from
+all the proxies i.e. we **won't** provide user ability to configure it just for
+proxy which raised those failures due to which ejection got triggered.
+
+Let me explain above piece with an example. Let's say there are two services A
+and B which calls C service. In case B->C gets success rate of 100% but A->C
+gets success rate of 50%, it won't be possible for user to configure
+service-mesh such that it would trigger circuit breaking only for A->C traffic.
+Instead either circuit breaking will happen for both A and B (through
+aggregation of their stats) or none of them.
+
+This certainly has its own disadvantages but I would be somewhat inclined
+towards having it simpler unless usecases for making it per-proxy over-weighs
+other one. So, for now, I propose that every ejection policy would be based on
+taking aggregated metrics over all the proxy instances to mark any instance for
+ejection.
+
+For HTTP/gRPC, I feel like taking advantage of linkerd provided golden metrics
+for ejection policies. Ejection can be based on success-rate,
+percentile-latencies for a given time window. Circuit breaker settings would be
+applied per service + protocol instead of per-route. In case circuit breaker is
+triggered for entire service, then we will return 503 for HTTP with some custom
+header and gRPC message for gRPC.
+
+If we go with the above design, may be we can keep most of the ejection logic in
+control plane i.e. in the destination API, we can just remove ejected endpoints
+from list given to discovery.
+
+Other than above mentioned details, here's the top level changes I could think
+of with this approach -
+
+1. Updating ServiceProfile CRD and its validation.
+2. Updating destination service to use circuit breaking configurations specified
+   by user to perform ejection. It can use prometheus metrics for evaluation and
+   update endpoints as required.
+3. Sharing some information through destination API to linkerd2-proxy so that it
+   can return adequate response to the requests.
+4. AFAIK metrics exposed by each proxy doesn't container upstream service
+   instance's endpoint.
+5. Exposing metrics related to circuit breaking from control plane.
+
+There's lot more things which needs to be discussed but I would first like to
+get some feedback if basic idea sounds sane to everyone.
+
+### Prior art
+
+[prior-art]: #prior-art
+
+Other service meshes seems to be performing this in data plane instead, may be
+because it aligned well with their architecture. All the service meshes built
+over top of enovy, uses envoy's inbuilt capability of circuit breaking. In
+[envoy](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/outlier),
+they use two terms - "Circuit Breaking" and "Outlier Detection", which I feel
+seems inter-related upto some extent and can be merged into single. Envoy
+supports both cluster-wide as well as per-proxy outlier detectors.
+
+With the above proposed design, we won't be able to support per-proxy outlier
+detectors cleanly, since we are abstracting almost all circuit-breaking logic
+from data plane. Also I guess getting metrics about other proxies in the cluster
+can potentially be a pain point if we move this logic into data plane.
+
+### Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+ToDo.
+
+### Future possibilities
+
+[future-possibilities]: #future-possibilities
+
+ToDo.
--- a/design/0005-circuit-breaking.md
+++ b/design/0005-circuit-breaking.md
@@ -30,14 +30,6 @@ rather than getting just another expected failure.
 
 ---
 
-Users should also be able to set max-ejection-percentage which provides max
-number of service instances which can be removed from the load balancing for
-specific service.
-
-[Example](https://drive.google.com/open?id=1jSMaogdvzxjv0v2dBoL7OYGHmsCusXn-)
-
----
-
 Network Parition - There can be lot of scenarios in network partition but here
 I'm providing a glimpse how circuit breaking would behave in case of network
 issue.
@@ -114,10 +106,6 @@ routes:
       # Circuit Breaker Kind
       # (Required)
       kind: ...
-
-      # Max percentage of instances that can be ejected.
-      # Default - 100%
-      maxEjectionPercentage: 10
 
       # The min duration of time in seconds for which an instance is marked
       # as ejected.
@@ -243,6 +231,11 @@ supports both cluster-wide as well as per-proxy outlier detectors.
    transport and network layers errors for now. That means any connection errors
    would be won't be triggering the circuit breaking. We can revisit the same in
    future.
+2. Users should also be able to set max-ejection-percentage which provides max
+   number of service instances which can be removed from the load balancing for
+   specific service. Since I don't have a good usecase in my mind, I'll keeping
+   it out of scope for now.
+   [Example](https://drive.google.com/open?id=1jSMaogdvzxjv0v2dBoL7OYGHmsCusXn-)
 
 ### Unresolved questions
 
--- a/design/0005-circuit-breaking.md
+++ b/design/0005-circuit-breaking.md
@@ -52,8 +52,17 @@ Protocol Specific features -
    User can configure ejection based on many different metrics. These metrics
    would be calculated over a given time range.
    - Success Rate
+     - Usecase: Most intuitive and generalised circuit breaking policy based on
+       failure/success classification selected by user.
    - Percentile Latencies
+     - Usecase: Let's say we have a service for which we cannot configure
+       timeouts due to any reason. For example, few of the requests are expected
+       to take longer than others. In this case, we cannot have circuit breaking
+       on success rate. In these cases, we can rely on latency based circuit
+       breaking
    - Consecutive Failures
+     - Usecase: In case, there are lot of failing requests due to some issue, we
+       need to fail quickly instead of waiting for next time window evaluation
 2. For TCP based communication -
    [Example](https://drive.google.com/open?id=1yAAurygsZQ-r58TEA0QiFf9Sz15jqRJV)
 
--- a/design/0005-circuit-breaking.md
+++ b/design/0005-circuit-breaking.md
@@ -38,9 +38,14 @@ specific service.
 
 ---
 
-Network Parition - There can be lot of scenarios in network partition but to
-provide a glimpse how circuit breaking would behave in case of network issue -
+Network Parition - There can be lot of scenarios in network partition but here
+I'm providing a glimpse how circuit breaking would behave in case of network
+issue.
+
 [Example](https://drive.google.com/open?id=154I2ZGlGc-rQJ0YVeek17dJy7raAvd-G)
+
+We won't be handling any errors in the network layer. Please check
+[out-of-scope section](#out-of-scope) for more details.
 
 ---
 
@@ -63,7 +68,9 @@ Protocol Specific features -
    - Consecutive Failures
      - Usecase: In case, there are lot of failing requests due to some issue, we
        need to fail quickly instead of waiting for next time window evaluation
-2. For TCP based communication -
+2. For TCP based communication (We won't be handling TCP related errors in the
+   implementation, please check [out-of-scope section](#out-of-scope) for more
+   details -
    [Example](https://drive.google.com/open?id=1yAAurygsZQ-r58TEA0QiFf9Sz15jqRJV)
 
 ---
@@ -86,9 +93,6 @@ current architecture, I feel it's best to provide circuit breaking functionality
 through service profiles. That means it would be limited to above two kind of
 services as of now. Circuit breaking logic would be applied in the data plane
 based on the metrics recorded inside the proxy.
-
-Since circuit breaking needs to be handled per protocol, we will leave TCP for
-now and can revisit the same in future.
 
 In future, for any other protocol support for circuit breaking, one will first
 need to incorporate the protocol into ServiceProfile itself and then implement
@@ -230,6 +234,15 @@ of circuit breaking. In
 they use two terms - "Circuit Breaking" and "Outlier Detection", which I feel
 seems inter-related upto some extent and can be merged into single. Envoy
 supports both cluster-wide as well as per-proxy outlier detectors.
+
+### Out of Scope
+
+[out-of-scope]: #out-of-scope
+
+1. Since circuit breaking needs to be handled per protocol, we will leave
+   transport and network layers errors for now. That means any connection errors
+   would be won't be triggering the circuit breaking. We can revisit the same in
+   future.
 
 ### Unresolved questions
 
--- a/design/0005-circuit-breaking.md
+++ b/design/0005-circuit-breaking.md
@@ -30,13 +30,6 @@ rather than getting just another expected failure.
 
 ---
 
-It would also be somewhat helpful in case of canary deployments when there's
-some regression and new build is raising failures.
-
-[Example](https://drive.google.com/open?id=1FxKi-egyhHumCvlHvTI4P114z7qOLPeN)
-
----
-
 Users should also be able to set max-ejection-percentage which provides max
 number of service instances which can be removed from the load balancing for
 specific service.
--- a/design/0005-circuit-breaking.md
+++ b/design/0005-circuit-breaking.md
@@ -53,7 +53,6 @@ Protocol Specific features -
    would be calculated over a given time range.
    - Success Rate
    - Percentile Latencies
-   - Max Requests
    - Consecutive Failures
 2. For TCP based communication -
    [Example](https://drive.google.com/open?id=1yAAurygsZQ-r58TEA0QiFf9Sz15jqRJV)
@@ -126,12 +125,6 @@ kind: successRate
 # (Required)
 timeWindow: 15
 successRate: 99
-```
-
-```yaml
-kind: maxRequests
-timeWindow: 15
-maxRequests: 10000
 ```
 
 ```yaml
--- a/design/0005-circuit-breaking.md
+++ b/design/0005-circuit-breaking.md
@@ -179,6 +179,24 @@ pair needs to be marked as ejected or if any instance has served ejection time
 and it's ready to be brought back online. In case it's consecutive ejection,
 increase ejectionTime according to `ejectionTimeInflation` settings.
 
+```txt
+Window based checking -
+
+0 . . . . 10 . . . . 20 . . . . 30
+
+|<------->|<-------->|<--------->|
+C         C          C           C  <- Background thread checking stats after
+                                       every x seconds and see if there's
+                                       need to circuit break.
+```
+
+Additional metrics we need to add:
+
+1. Currently we can't get metrics for last `x` seconds. They are stored as
+   global counters.
+2. We need to add metrics for consecutive failures.
+   [Ref](https://github.com/linkerd/linkerd2-proxy/blob/5264573433ceea37f9d66c9ca95c458a604350a0/linkerd/http-metrics/src/requests/mod.rs#L15-L46)
+
 ---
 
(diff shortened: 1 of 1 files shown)


design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,150 @@
...
+Instead either circuit breaking will happen for both A and B (through
+aggregation of their stats) or none of them.
+
+This certainly has its own disadvantages but I would be somewhat inclined
+towards having it simpler unless usecases for making it per-proxy over-weighs
+other one. So, for now, I propose that every ejection policy would be based on
+taking aggregated metrics over all the proxy instances to mark any instance for
+ejection.
+
+For HTTP/gRPC, I feel like taking advantage of linkerd provided golden metrics
```
grampelberg: Why wouldn't we want to do circuit-breaking on a per-route basis?

design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,150 @@
...
+for ejection policies. Ejection can be based on success-rate,
+percentile-latencies for a given time window. Circuit breaker settings would be
+applied per service + protocol instead of per-route. In case circuit breaker is
+triggered for entire service, then we will return 503 for HTTP with some custom
+header and gRPC message for gRPC.
+
+If we go with the above design, may be we can keep most of the ejection logic in
```
grampelberg: It is likely that each proxy will have a different view of the world (imagine a network split). Wouldn't it be better to have the logic inside each proxy, as they'll be the only ones that have a full picture of what's going on?

design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,150 @@
...
+control plane i.e. in the destination API, we can just remove ejected endpoints
+from list given to discovery.
+
+Other than above mentioned details, here's the top level changes I could think
+of with this approach -
+
+1. Updating ServiceProfile CRD and its validation.
```
grampelberg, line 206: Mind adding some sample YAML with the changes in it?

design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,150 @@
...
+[design-proposal]: #design-proposal
+
+Right now if I just consider ServiceProfile spec, IMO it seems to represent more
+of profile-for-http-service rather than profile-for-generic-service. May be it
+makes sense given that it supports only HTTP and grpc over HTTP as of now. Given
+current architecture, I feel it's best to provide circuit breaking functionality
+through service profiles. That means it would be limited to above two kind of
+services as of now.
+
+For TCP connections, according to original github
```
grampelberg: TCP feels a little weird for circuit breaking. Maybe we should keep it out of scope for now? That'll help focus this a little better. To your earlier point, it feels like circuit breaking will need to be sorted out on a per-protocol level.

design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,150 @@
...
+
+[summary]: #summary
+
+We would like to implement circuit breaking functionality which would allow end
+users to configure mesh such that they can limit the impact of failures from
+faulty service over others and prevent cascading issues. By configuring circuit
+breaking, they will be able to remove failing service instances from load
+balancing and survive through it in a more graceful manner.
+
+## Problem Statement
```
grampelberg, line 21: It is super helpful in the problem statement to be very explicit about the concrete use case you're imagining. Would you mind providing a diagram with all the services, their behavior and walk through step by step for what should happen? This'll really help prioritize what features we'd like to see in the design section.

design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,150 @@
...
+I'm not sure of its behaviour -- whether TCP error means connection refused or
+resets or timeouts or something else. Anyway, it seems like it's not
+configurable and I'm not sure if making it configurable has any important
+usecases.
+
+In future, for any other protocol support for circuit breaking, one will first
+need to incorporate the protocol into ServiceProfile itself and then implement
+the circuit breaking functionality.
+
+Regarding ejection of the service instance, once we decide any instance needs to
```
grampelberg: Mind going into this a little more? I'm not sure what you're trying to say.

design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,150 @@
...
+
+[prior-art]: #prior-art
+
+Other service meshes seems to be performing this in data plane instead, may be
+because it aligned well with their architecture. All the service meshes built
+over top of enovy, uses envoy's inbuilt capability of circuit breaking. In
+[envoy](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/outlier),
+they use two terms - "Circuit Breaking" and "Outlier Detection", which I feel
+seems inter-related upto some extent and can be merged into single. Envoy
+supports both cluster-wide as well as per-proxy outlier detectors.
```
grampelberg, line 224: AFAIK "cluster-wide" in Envoy parlance just means a configuration group right? You'd need to implement an xDS controller that consumes per-envoy metrics to orchestrate multiple envoy proxies?

design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,150 @@
...
+get some feedback if basic idea sounds sane to everyone.
+
+### Prior art
+
+[prior-art]: #prior-art
+
+Other service meshes seems to be performing this in data plane instead, may be
+because it aligned well with their architecture. All the service meshes built
+over top of enovy, uses envoy's inbuilt capability of circuit breaking. In
+[envoy](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/outlier),
```
grampelberg, line 221: Reading through the Envoy docs, it seems that they suggest you use retries and `max_retries` to implement latency based circuit breaking?

grampelberg (commented):
Sorry this took me so long! I asked a bunch of questions, would love to know what you think =)

design/0005-circuit-breaking.md
Ashish-Bansal: Yes, envoy can somewhat emulate the behaviour of latency based circuit breaking based on `max_retries`. 

`max_retries` - The maximum number of parallel retries that Envoy will allow to the upstream cluster.

In case a specific route has retries configured and due to some fault, requests starts failing due to which no. of parallel retries increases more than the `max_retries`, then circuit breaker will be triggered and circuit will be in 'open' state. 

But we can't configure the same for the non-idempotent requests. So, it's still a limitation. 

--- 

Though it's still possible to have "Outlier Detection" mechanism for non-idempotent requests. For example, instead of configuring retries, I can just configure timeout for those routes. Along with those timeouts, I can add configuration for "Outlier detection" based on success rate. So, in case request's response time increases beyond timeout, they would be marked as 'failed requests' and it will decrease success rate which will trigger [outlier-based-circuit-breaking](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/outlier#success-rate).

design/0005-circuit-breaking.md
Ashish-Bansal: This explains what happens in cluster-wide and per-proxy outlier detection according to Envoy's definition. 

image

> You'd need to implement an xDS controller that consumes per-envoy metrics to orchestrate multiple envoy proxies?

image

You mean something like above ? No envoy doesn't support that. Along with that, I think it won't be super beneficial given that every downstream proxy instance would eventually realise that upstream instances are causing failures and they would 'individually' circuit break them.

design/0005-circuit-breaking.md
Ashish-Bansal: > it feels like circuit breaking will need to be sorted out on a per-protocol level.

Hm, right. Yes, it's safe to drop TCP for now in that sense.

design/0005-circuit-breaking.md
Ashish-Bansal: tldr; this would be invalid given that I completely agree with moving the circuit breaker logic to the data plane. 

I have attached the diagram regarding what I meant earlier. Let's say there are two services A and B which calls C service. In case B->C gets success rate of 100% but A->C gets success rate of 0%, it won't be possible for user to configure service-mesh such that it would trigger circuit breaking only for A->C traffic. So, either circuit breaking will happen for both A and B (through aggregation of their stats) or none of them.

image

design/0005-circuit-breaking.md
Ashish-Bansal: It was more of a drawback with having circuit breaking logic in the control plane rather than the data plane. 

By having all this logic in the data plane, nothing stops us from allowing user to have circuit-breaking for both per-route and per-service basis.

And yes, per-route basis circuit breaking would allow more fine grained control and would be helpful in many cases.

design/0005-circuit-breaking.md
Ashish-Bansal: Hm, that's true. Yes, it certainly makes sense to have this in data plane. I'm not sure what I was thinking while writing it, it was plain stupid.

Ashish-Bansal (commented):
Thanks for the review @grampelberg!

I'll update the proposal with mentioned changes.

hodglem:
Hi. I asked a question in the email group about circuit breaking, rate limiting and timeouts for Linkerd 2. Charles said I should post here about our use case and thoughts. Thanks for taking the time to listen :)

We are doing a POC using Linkerd2 and one of the benefits that we were hoping to get out of adopting SM was circuit breaking. As an example, one deployment has many pods. Linkerd balances the traffic to those pods equally today in a stable state. Linkerd also seems to balance the traffic slowly when deployments are scaled up to ensure that the new pods don't puke. Both really cool features.

Assuming a stable state of the above example, in the case where one pod starts to fail, we'd like to be able to configure Linkerd (through Service Profiles I guess?) to, at minimum, reroute that traffic to healthy pods. A bigger ask is to have Linkerd work with Flagger (or something) to scale deployments based on some condition. Overload protection comes to mind here. Configurable Rate limiting and Timeout rules would be nice to have also.

Looking at Linkerd 1, Failure Accural seems to do most of what we'd want and if that was to work in Li nkerd2, that would probably suffice but it sounds like you'd like a more complete solution. Also, and I'm guessing, but it seems like you guys moved away from the Config scheme in Linkerd 1 and have an entirely new method via service profiles, so it probably isn't as easy as just bringing that feature over.

(For some context, I have been using [this book](https://www.packtpub.com/web-development/mastering-service-mesh-architecture) as  a reference. Khatri mentions "Automatic Scaling" and "Resliency" as key features of Servcie Mesh. As mentioned in https://github.com/linkerd/linkerd2/issues/2846, other providers have this kind of behavior.)

I hope this is helpful. Let me know if you folks have questions.

design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,231 @@
...
+
+Users should also be able to set max-ejection-percentage which provides max
+number of service instances which can be removed from the load balancing for
+specific service.
+
+[Example](https://drive.google.com/open?id=1jSMaogdvzxjv0v2dBoL7OYGHmsCusXn-)
+
+---
+
+Network Parition - There can be lot of scenarios in network partition but to
```
grampelberg: Are you saying that circuit breaking should *only* be for success rate and not have anything to do with connection errors?

design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,231 @@
...
+---
+
+Protocol Specific features -
+
+1. For HTTP/gRPC based communication, users should be able to configure circuit
+   breaking either per endpoint or whole service.
+   [Example](https://drive.google.com/open?id=1JK65u9Mnmkmy_X8fczeh6RO6O-vNiTtY)
+   User can configure ejection based on many different metrics. These metrics
+   would be calculated over a given time range.
+   - Success Rate
```
grampelberg, line 51: I think this generally makes sense, but which use case do each of these support?

design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,231 @@
...
+   - Percentile Latencies
+   - Max Requests
+   - Consecutive Failures
+2. For TCP based communication -
```
grampelberg: I got the impression from above that circuit breaking shouldn't take connectivity errors into account. This seems to contradict that?

design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,231 @@
...
+
+Right now if I just consider ServiceProfile spec, IMO it seems to represent more
+of profile-for-http-service rather than profile-for-generic-service. May be it
+makes sense given that it supports only HTTP and grpc over HTTP as of now. Given
+current architecture, I feel it's best to provide circuit breaking functionality
+through service profiles. That means it would be limited to above two kind of
+services as of now. Circuit breaking logic would be applied in the data plane
+based on the metrics recorded inside the proxy.
+
+Since circuit breaking needs to be handled per protocol, we will leave TCP for
```
grampelberg: Let's create a "out of scope" section to track the things we're explicitly not working on yet.

design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,231 @@
...
+now and can revisit the same in future.
+
+In future, for any other protocol support for circuit breaking, one will first
+need to incorporate the protocol into ServiceProfile itself and then implement
+the circuit breaking functionality.
+
+For HTTP/gRPC, we already have golden metrics available in the proxy. Along with
+that, we can add additional metrics and use those for circuit breaking logic. In
```
grampelberg, line 94: What additional metrics would we need to add?

design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,231 @@
...
+---
+
+It would also be somewhat helpful in case of canary deployments when there's
+some regression and new build is raising failures.
+
+[Example](https://drive.google.com/open?id=1FxKi-egyhHumCvlHvTI4P114z7qOLPeN)
+
+---
+
+Users should also be able to set max-ejection-percentage which provides max
```
grampelberg: Why is this important? What is the use case?

design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,231 @@
...
+service is unable to respond due to any reason, it's possible that these
+failures can cascade to other services in the cluster.
+
+If we are aware that specific service instance is not responding, we can call it
+unhealthy and it's better if we just stop sending "traffic"[0] to that instance
+rather than getting just another expected failure.
+
+---
+
+It would also be somewhat helpful in case of canary deployments when there's
```
grampelberg: For canary, wouldn't you just shift the load back? Why would you want a circuit breaker specifically?

design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,231 @@
...
+## Problem Statement
+
+In a service mesh, services often make calls to other services. In case upstream
+service is unable to respond due to any reason, it's possible that these
+failures can cascade to other services in the cluster.
+
+If we are aware that specific service instance is not responding, we can call it
+unhealthy and it's better if we just stop sending "traffic"[0] to that instance
+rather than getting just another expected failure.
+
```
grampelberg, line 30: Isn't there a use case to quick fail a misbehaving service (imagine having a db go down)?

design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,231 @@
...
+   User can configure ejection based on many different metrics. These metrics
+   would be calculated over a given time range.
+   - Success Rate
+   - Percentile Latencies
+   - Max Requests
+   - Consecutive Failures
+2. For TCP based communication -
+   [Example](https://drive.google.com/open?id=1yAAurygsZQ-r58TEA0QiFf9Sz15jqRJV)
+
+---
```
grampelberg, line 68: Once the circuit is broken, what should be returned to the client? In all these cases a client should get some kind of notification. If we're removing unhealthy instances, that might cause a thundering herd on everything else and take the entire cluster down.

Is there any kind of propagation that should happen between services?

design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,231 @@
...
+      kind: ...
+
+      # Max percentage of instances that can be ejected.
+      # Default - 100%
+      maxEjectionPercentage: 10
+
+      # The min duration of time in seconds for which an instance is marked
+      # as ejected.
+      # (Required)
+      baseEjectionTime: 60
```
grampelberg, line 113: Why does this needs to be configurable?

design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,231 @@
...
+
+      # Ejection Time Inflation
+      # How to increase ejection time on every consecutive ejection.
+      # (Optional. Default - None)
+      ejectionTimeInflation: ...
```
grampelberg, line 118: Shouldn't we just use the exponential backoff algorithm that k8s uses? Why would someone need to configure this? What use case does it support?

design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,231 @@
...
+routes:
+  - name: "/foo"
+    circuitBreaker:
+      # Circuit Breaker Kind
+      # (Required)
+      kind: ...
+
+      # Max percentage of instances that can be ejected.
+      # Default - 100%
+      maxEjectionPercentage: 10
```
grampelberg: Should we be thinking of weights instead of percentages?

design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,231 @@
...
+
+```yaml
+kind: percentileLatency
+timeWindow: 15
+# p99 or p95 or p50 in ms
+p99: 500
+```
+
+```yaml
+kind: consecutiveFailures
```
grampelberg, line 139: Why is this different than `successRate`?

design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,231 @@
...
+ejectionTimeInflation:
+  # Increases ejection time exponentially based on provided multiplier
+  kind: exponential
+  multiplier: 2
+  maxEjectionTime: 120
+```
+
+---
+
+Circuit breaking logic -
```
grampelberg, line 167: How does this fit with latency based load balancing and traffic split weights?

design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,231 @@
...
+
+---
+
+Other than above mentioned details, here's the top level changes I could think
+of with this approach -
+
+1. Updating ServiceProfile CRD and its validation.
+2. Updating destination service APIs to provide circuit breaking configuration
+   to proxy instances.
+3. Using existing local telemetry to apply circuit breaking configuration in
```
grampelberg: It sounds like this means there wouldn't be a dependency on prometheus. Do we need to track histograms and the like internally in the proxy then?

design/0005-circuit-breaking.md
```diff
@@ -0,0 +1,231 @@
...
+   proxy itself.
+
+All diagrams are collectively available
```
grampelberg, line 212: This is super awesome and helpful, thank you!!!

design/0005-circuit-breaking.md
Ashish-Bansal: Hm yeah, I think that could be handled using `Consecutive Failures`. In case db is down, *most* of the requests would be failing and can be identified through this.

Ashish-Bansal: Not to mention, they can be identified through other kind of circuit breakers as well like 'success rate' but since those ones use moving window, they can potentially cause issue during that 'window period'.

design/0005-circuit-breaking.md
Ashish-Bansal: Hm, I'm not fully aware of solutions like flagger and how they handle canary deployments. 

> For canary, wouldn't you just shift the load back

But yes, general idea would be shifting back the load. In case, there's circuit breaker configured, new instances would be circuit break-ed and traffic would be sent only to old instances i.e. in a way we will be shifting the load back. Though yeah, I'm not fully sure how it would fit well with other available tools in a bigger picture.

design/0005-circuit-breaking.md
Ashish-Bansal: > not have anything to do with connection errors

In case of network errors i.e. we are not able to reach the host itself, yes it won't involve circuit breaking. It's similar to how we are handling things in case of TCP. 

In case we do plan to handle circuit breaking for network errors or at any other layer, it will be handled differently. Here we will just be handling issues arising from the HTTP layer (or gRPC etc). Anything happening below application layer, nope, we won't be handling it.

design/0005-circuit-breaking.md
Ashish-Bansal: Ah yes, I don't plan to implement this functionality.

Ashish-Bansal: Here I just mean to add the features which a circuit-breaker may have and we can select ones which we want to add, to make sure we are not overlooking anything.

design/0005-circuit-breaking.md
Ashish-Bansal: > Once the circuit is broken, what should be returned to the client? In all these cases a client should get some kind of notification.

Response sent to the client depends on kind of request. In case of HTTP, it would be 503 with some custom header in the response. For gRPC, it will be some kind of gRPC message.

design/0005-circuit-breaking.md
Ashish-Bansal: Yep, makes sense! (y) 

ToDo - add "out of scope" section.

design/0005-circuit-breaking.md
Ashish-Bansal: `successRate` calculates over the time period of window. 

For example, if time window is 15 seconds, we will calculate success rate using no. of requests which got successful response within last 15 seconds / total no. of requests within those 15 seconds. 

But for consecutiveFailures, I can set in case `x` no. consecutive requests got failure, just eject the upstream instance immediately. It won't have to wait for the large no. of request to fail so that success rate circuit breaker gets triggered.

design/0005-circuit-breaking.md
Ashish-Bansal: > It sounds like this means there wouldn't be a dependency on prometheus

Right.

> Do we need to track histograms and the like internally in the proxy then?

Yes, we will need to track all those stuff internally inside the proxy but I haven't thought carefully about the way and data structure we would be using to save the data inside proxy.

design/0005-circuit-breaking.md
Ashish-Bansal: Hm, before making a comment, I'll need to look into proxy to see what kind of metrics and how we are saving currently for exposing them to prometheus.

design/0005-circuit-breaking.md
Ashish-Bansal: Initially I was having concerns about its impact on memory usage of the proxy but after looking into the envoy, I found that it's storing them locally as well, so I ruled them out. But yes, I think it's important to clarify what all we would be storing inside proxy.

design/0005-circuit-breaking.md
Ashish-Bansal: I feel it should be customisable by the user so that they can set the ejection period according to the context -- the kind of load, requests they have. In some cases, they may want the circuit breaker to get to into half-open state quickly just due to flakiness they have in their architecture and in some cases, they may want it to be high so that frequent half-open state doesn't result into failure cascades.

design/0005-circuit-breaking.md
Ashish-Bansal: > Why would someone need to configure this? What use case does it support?

Same as above. Though I understand that having lot of switches can be confusing for the end user but we can add them up with sane defaults so that it's win-win situation for both non-advanced and fine tuners.

Ashish-Bansal: But I'm fine with keeping it simple as well and add it in future if need be.

design/0005-circuit-breaking.md
Ashish-Bansal: Hm, I got it from the envoy and it was more of a, ahem ahem, blind copying. 

To think about it again, it means we will keep sending traffic to the upstream even when the circuit breaker should be in open state for those upstreams. So, in a way, it can act as a safe default. For example, let's say max-ejection-percentage is 10%, user can modify all other configuration options related to circuit-breaker and see how those settings affect the cluster in different scenarios. This way they won't end up taking down their whole service due to the same. But yeah, I'm not sure if it's convincing enough to have it for such a usecase -- well it's not for my own personal use case.

design/0005-circuit-breaking.md
Ashish-Bansal: 1. Currently the way we are storing metrics, we can't get metrics for a window of certain time period. They are stored as global counters.
2. We need to add metrics for consecutive failures.

Ref - https://github.com/linkerd/linkerd2-proxy/blob/5264573433ceea37f9d66c9ca95c458a604350a0/linkerd/http-metrics/src/requests/mod.rs#L15-L46

design/0005-circuit-breaking.md
Ashish-Bansal: About traffic splits, based on the SMI spec, we assign weights to the group of upstreams and then we divide the traffic between those groups based on weights.
Then latency based load balancing is applied based on the upstreams of a single group. Is that correct ?

For circuit breaking, if we will be ejecting the pods completely, we won't need to worry about traffic split and latency based load balancing because they will continue to distribute load same as before. Only difference is now the no. of upstreams would be lower.

design/0005-circuit-breaking.md
Ashish-Bansal: Yes, we will need to store those histograms internally for the last window. 

Ref - https://github.com/linkerd/rfc/pull/30/files#r464773186

Ashish-Bansal:
```
Window based checking -

0 . . . . 10 . . . . 20 . . . . 30

|<------->|<-------->|<--------->|
C         C          C           C  <- Background thread checking stats after
                                       every x seconds and see if there's
                                       need to circuit break.
```

We will only be storing it for the last window.

design/0005-circuit-breaking.md
Ashish-Bansal: ToDo - add `ejectAllPercentageMark` -

Eject all instances once x% of upstream are already ejected. 

Usecase explanation -- it will be helpful when user know remaining % of upstreams won't be able to handle the incoming traffic and they are going to eventually fail.

--- 

Let me know what you think about the same. In case it sounds sane, I'll add it to the doc.

design/0005-circuit-breaking.md
Ashish-Bansal: We are going to eject instance completely from the upstream instance rather than having decreasing weights for unhealthy instances. So, in that approach, having percentages is way to go. 

But as mentioned [here](https://github.com/linkerd/rfc/pull/30/files#r461794191), I haven't been able to provide convincing explanation for the same, so I'll move it to out of scope for now.

design/0005-circuit-breaking.md
Ashish-Bansal: Usecases - 

Success Rate - Most intuitive and generalised circuit breaking policy based on failure/success classification selected by user.

Latency based rate limiting - Let's say we have a service for which we cannot configure timeouts due to any reason. For example, few of the requests are expected to take longer than others. In this case, we cannot have circuit breaking on success rate. In these cases, we can rely on latency based circuit breaking.

Consecutive failures - In case, there are lot of failing requests due to some issue, we need to fail quickly instead of waiting for next time window evaluation.

Max requests - Hm, may be circuit breaking isn't an ideal place to add this. I'll remove this one.

olix0r:

spockz:
The failure-accrual mechanism in combination with response classification has been very successful for us when using Finagle. (It is the same mechanism as in Linkerd v1.x obviously.)

It appears as if we are considering new things, primarily using references to Envoy. Why is the rich heritage of Finagle/LinkerdV1 being ignored?

rverma-dev:
Just wondering are considering a usecase on circuit-breaking at local sidecar too?

Something like

E.g. ALB -> Linkerd Sidecar A -> Service A
                                      \ -> Service A fall back

This can help in doing circuit breaking at edge too? Although one of the consideration is that the pod has to remain up  even if it with linkerd sidecar only being running state.

cpretzer:
> Just wondering are considering a usecase on circuit-breaking at local sidecar too?
> 
> Something like
> 
> E.g. ALB -> Linkerd Sidecar A -> Service A
> \ -> Service A fall back
> 
> This can help in doing circuit breaking at edge too? Although one of the consideration is that the pod has to remain up even if it with linkerd sidecar only being running state.

@rverma-nsl There's a separate [failover project](https://github.com/alpeb/linkerd-failover) that might work for the scenario you've described. Please have a look and let us know if that is what you had in mind

sherifkayad:
Are there any plans to merge that RFC and work on it?

Ashish-Bansal:
@sherifkayad There's a lot of stuff that needs modification and discussion in the proposal. But as for me, I don't have any plans to drive the same.

wmorgan:
We're currently working on adding circuit breaking to Linkerd. A few more details [in the roadmap blog post](https://linkerd.io/2021/12/29/the-service-mesh-in-2022/).

sherifkayad:
@wmorgan that's really amazing news! Thanks for the heads-up.

Ashish-Bansal:
Thanks for implementing circuit breaking!

Ref - https://github.com/linkerd/linkerd2/issues/2846#issuecomment-1502026374!

PR #31, opened 2020-06-19
draft of multicluster phase 2 design

adleong:

### DIFF
--- a/design/0004-multicluster-phase-2.md
+++ b/design/0004-multicluster-phase-2.md
@@ -38,6 +38,7 @@ I propose the following changes:
   the following annotations will be eliminated:
   * mirror.linkerd.io/gateway-name
   * mirror.linkerd.io/gateway-ns
+  
   This also eliminates arbitary constraint that an exported service needed to be
   fronted by a single gateway.  In reality, the target cluster may have any
   number of gateways and each of these has the ability to route to any service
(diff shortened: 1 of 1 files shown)


alpeb:
Have you considered using a CRD to handle each "link" against a target cluster? Would provide us easy config validation, and target clusters properties would become queriable through `kubectl`.
We could also go as far as handling link creation/teardown and overall lifecycle through webhooks...
Thoughts?

adleong:
@alpeb Thanks for bringing this up, it's a good idea to consider.

Can you elaborate a bit more on how handing lifecycle through webhooks would look?

After my experience with ServiceProfiles, I have to say I'm not a huge fan of CRDs.  client-go handles them using generated code which I've found to be very brittle and extremely difficult to version.  Making any changes to the CRD makes it extremely difficult or in some cases impossible to upgrade gracefully.

In this case specifically, this config is consumed only by the service-mirror-controller and is not involved in any coordination between components.  This means that we can have the service-mirror-controller be responsible for parsing and validation.  Since we're in total control the parsing and validation logic, we can have it be flexible enough to gracefully handle upgrades and changes to the schema.

alpeb:
I was thinking more on simply having a CRD without the API aggregation, so no need to get involved with the client-go autogen code. The controller would still need to be something like our current service-mirror controller, that would subscribe as a webhook listener to all the CRUD events happening to the CRD instances. Upon the CRD creation event the endpoints and probing would need to be setup just like in Proposal 1 in this RFC. Upon the CRD deletion event the endpoints would be torn down (btw what is the proposed mechanism for tear down in Proposal 1?). As in the current proposal, using CRD alleviates us from having to deal with watchers. But we could still not use webhooks and keep on using watchers. Also note this implies sticking with the current single-controller approach.

The main advantage in my view of a CRD vs a ConfigMap is that changes made to it are immediately validated; if the user makes a mistake the resource will be rejected upon applying it, whereas errors in the ConfigMap will have to be surfaced elsewhere. 

As for versioning, I'm out of the loop regarding our current difficulties, and whether that's relevant only to API aggregation. We can have a very minimal OpenAPI schema for the CRD and the service mirror controller (or a separate controller) can act as a validator webhook that we can evolve as needed.

An then `kubectl` can query these resources, and have `kubectl get` have columns showing things like the target cluster name, the service selector, etc.

grampelberg:
I like using CRDs when start having more than one piece of data that requires configuration. As we'll need a couple here, feels like a good improvement.

adleong:
Having instant validation via a validating webhook and having more useful data in `kubectl get` are definitely compelling reasons to use a CRD for this.

It looks like the service mirror controller could probably use the [dynamic client](https://godoc.org/k8s.io/client-go/dynamic) package to get the custom resource as an `Unstructured`.  Is that what you had in mind, @alpeb?

I still would prefer to have a separate service mirror controller to handle each target cluster.  But it could certainly load its configuration from a custom resource instead of a confgmap.

To answer your question about link teardown: tear down involves deleting all of the resources installed by `link` (service mirror controller, config, credentials secret, and RBAC) as well as mirror resources (mirror services, mirror endpoints, gateway mirror, gateway mirror endpoints).  If we use a label on all of these resources, it should be fairly straightforward:

```
linkerd --context=source delete -A (huge list of resource types) -l multicluster.linkerd.io/remote-cluster=target
```

But this is a tricky cumbersome command that includes a huge list of resource types.  To make this easier, we could bundle this functionality into an `unlink` command which outputs manifests to delete, just like `linkerd uninstall` does.

```
linkerd --context=source multicluster unlink --cluster-name=target | kubectl --context=source delete -f -
```

alpeb:
> It looks like the service mirror controller could probably use the [dynamic client](https://godoc.org/k8s.io/client-go/dynamic) package to get the custom resource as an `Unstructured`. Is that what you had in mind, @alpeb?

Requests to webhooks are raw bytes, so I was thinking on simply marshalling that with some ad-hoc struct, but `Unstructured` might be cleaner indeed.

design/0004-multicluster-phase-2.md
```diff
@@ -199,34 +201,62 @@ manifests for all of them:
...
\ No newline at end of file
+## Unlinking
+
+To unlink a target cluster, we add an `unlink` command which outputs manifests
+for the `link` resource, the service mirror controller deployment, the cluster
+credentials, and all of the mirror services and gateway mirror for the given
+target cluster.  These can be piped to `kubectl delete`.
+
+```
+linkerd --context=source multicluster unlink --cluster-name=foobar | kubectl delete -f -
```
olix0r, line 212: What's the behavior if no name is specified?

Are all clusters unlinked? That would be bad 'default' behavior...

Should this require a list of cluster names, e.g. `ulink foobar`?

design/0004-multicluster-phase-2.md
```diff
@@ -0,0 +1,262 @@
...
+kubectl --context=source get links
+NAME   NUM SVC   ALIVE   LATENCY_P50   LATENCY_P95   LATENCY_P99
+foobar       3    True         275ms         385ms         397ms
+```
+
+
+## Upgrading
+
+To upgrade the service mirror controller, simply run the link step again with a
+newer version of the CLI:
```
olix0r, line 190: What are the dependencies _between_ clusters? I.e. if i've got a gateway (target) cluster at version N and a mirror (source) cluster at version M, what are the things that must be common to both versions? Could these change? How will they change?

I also think this is oversimplified. How do you upgrade the CRD, for instance? Can't we "break" existing mirror configurations with changes to this CRD?

design/0004-multicluster-phase-2.md
```diff
@@ -0,0 +1,262 @@
...
+
+```
+linkerd --context=target multicluster link --cluster-name=foobar | kubectl --context=source apply -f -
+```
+
+Alternatively, we can add an `update` subcommand to `multicluster` which finds
```
olix0r, line 196: upgrade?

design/0004-multicluster-phase-2.md
adleong: Just as in the `link` command, the `--cluster-name` flag is mandatory.  

We also describe an explicit `--all-clusters` flag below.

design/0004-multicluster-phase-2.md
adleong: It's hard to speculate about dependencies between clusters without knowing in advance of WHAT will change.  For example, the `link` command extracts certain information from the target cluster such as the gateway address and the service account token.  If the location or format of this information has changed, this will not work.  

I suspect that you would have to upgrade the target cluster first in this case by running `linkerd mc install` before running `link` again.  But I'm not sure this speculation is very useful without knowing in advance what is changing.

We avoid using generated code to interact with the CRD specifically to avoid issues during upgrades.  The service mirror controller will read the `link` resource using a dynamic client that gets the contents as an untyped map.  This allows us to build in our own backwards compatibility and ensure that if the format of the CRD changes that the service mirror controller is able to handle both the old and new format.

design/0004-multicluster-phase-2.md
adleong: yes, thank you

design/0004-multicluster-phase-2.md
```diff
@@ -0,0 +1,262 @@
...
+source cluster.
+
+Active links can be seen using `kubectl get`.  This replaces the `multicluster gateways` command:
+
+
+```
+kubectl --context=source get links
+NAME   NUM SVC   ALIVE   LATENCY_P50   LATENCY_P95   LATENCY_P99
+foobar       3    True         275ms         385ms         397ms
+```
```
alpeb, line 184: This might delve a little into the implementation design phase, but I just wanted to dig a little bit and throw some high level ideas on how this could be addressed...

As I understand it, extra printer columns are defined so that they show the values of some properties of the CRD instance, retrieved through a jsonpath expression. Stuff like the target cluster name and the services selector are readily available. But metrics like latencies would need to be actively surfaced into the CRD instance, as status properties I guess (is this what the [status subresource](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#status-subresource) is for?). The probe worker itself could update that status as it updates the latencies prometheus metrics.

olix0r:

PR #32, opened 2020-06-24
Support Minimal Control Plane install

Pothulapati:
https://github.com/linkerd/linkerd2/issues/4612

_This RFC aims to get high level information, suggestions, questions, feedback on the minimal control-plane installation, so that the implementation can be drafted accordingly._

## Summary

[summary]: #summary

An incremental approach to installing Linkerd, with support for additional dependencies.

## Problem Statement (Step 1)

Linkerd's control plane components can work independently(example: Tap, etc)
but currently, the installation logic is tightly coupled i.e in a single chart
(except for prometheus and grafana). This has the following problems:

- Components can't be skipped during installation.
- No incremental way of installing components as required.

Once the proposal is implemented, Users will be able to

- Use the Linkerd CLI or Helm to install a minimal control plane with only the
bare essentials of functionality.
- Use the Linkerd CLI or Helm to install addon charts such as Linkerd-viz to add
functionality to the minimal control plane independently.
- Use `linkerd install` or full control-plane to install a complete Linkerd control
plane including full observability, just like they can today.

### Goals

- Remove tight coupling between control-plane components with installation.
- Provide an incremental installation approach of the control-plane along with a
upgrade story.
- Update Dashboards and CLI to work as per the installed components.

[Rendered](https://github.com/linkerd/rfc/blob/tarun/min-cp/problem/0006-minimal-controlplane-install.md)

cc: @grampelberg


olix0r:

PR #33, opened 2020-07-21
Design for annotation-configured mTLS for server-first protocols

olix0r:


design/0005-protocol-detection.md
```diff
@@ -0,0 +1,308 @@
...
+must be preserved and transmitted to the remote proxy so that it can route it
+properly. This is transmitted via a new _Linkerd Connection Header_.
+
+#### Multi-Cluster Service Mirror Controller
+
+The service-mirror creates local endpoints that reflect the opaqueness, as
+configured in the target cluster. The multi-cluster gateway then uses the
+connection header to route the connection through the gateway.
+
+### Proxy handling of inbound-targeted connections
```
zaharidichev, line 129: Just so I make sure I understand correctly... The search for ` _Linkerd Connection Header_`  is happening only when the inbound connection is happening on the proxy inbound port directly?

design/0005-protocol-detection.md
```diff
@@ -0,0 +1,308 @@
...
+#### Linkerd Connection Header preamble
+
+The connection header is a preamble that is transmitted from outbound proxies
+to remote inbound proxies. This preamble consists of:
+
+* a 12B magic marker;
+* a 4B message length; and
+* a variable-length protobuf-encoded message.
+
+When the first 16B of a stream match the magic marker, the proxy strips the
```
zaharidichev, line 154: When the first 16B -> When the first 12B ?

design/0005-protocol-detection.md
```diff
@@ -0,0 +1,308 @@
...
+`0x6c35642e696f2f70726f7879`.
+
+Following that, a 4-byte unsigned integer (Big Endian) message length,
+followed by a protobuf-encoded `LinkerdConnectionHeaderV1` message.
+
+This message type includes at least an optional application port, and an
+optional detection hint.
+
+When the port is set, it identifies a port on localhost to which the traffic
+should be forwarded. When unset (`0`) or set to the value of the inbound or
```
zaharidichev, line 167: If we are using a proto structure, I assume this can actually be expressed as an Optional val (I think), rather thansetting it to 0. WDYT ?

olix0r: When using a simple value like
```proto3
int port = 1;
```
the default value is 0.

If you wanted to have it manifest as an option, you have to do something like:
```proto3
message Port {
  int port = 1;
}
Port port = 1;
```

I don't think there's any real value in the compexity.  We have to handle a 0 value either way.

design/0005-protocol-detection.md
olix0r: The RFC shouldn't specify that, imo. I'm only trying to define the behavior for what happens when a connection targets the inbound port. It's an implementation detail as to whether we attempt this on other connections...

design/0005-protocol-detection.md
```diff
@@ -0,0 +1,308 @@
...
+outbound proxy ports, the proxy handles the connection as it would a gateway
+connection. I.e., the proxy r when configured as a multi-cluster gateway)
```
olix0r, line 169: ...

design/0005-protocol-detection.md
olix0r, line 154: 12

design/0005-protocol-detection.md
```diff
@@ -0,0 +1,308 @@
...
+
+* a 12B magic marker;
+* a 4B message length; and
+* a variable-length protobuf-encoded message.
+
+When the first 16B of a stream match the magic marker, the proxy strips the
+entire header from the connection before proxying it to the application.
+
+These messaged are prefaced with a magic marker prefix: `l5d.io/proxy`, or
+`0x6c35642e696f2f70726f7879`.
```
olix0r, line 158: this should probably end in `\r\n\r\n` newline so that we cannot possibly be mistaken for an http connection

olix0r:

PR #34, opened 2020-08-07
Design Proposal - Implementing SMI Traffic Specs and Enhancing Traffic Splits

campbel:


olix0r:

rbtcollins:
Further to the current prose about deployments I'd like to add that additional features such as injected developer builds are pretty much impossible without some form of tagged traffic route rather than pure weighted stochastic routing. Keeping the surface area clean for developer UX is important, and why we're still evaluating LinkerD, but this feature is very interesting to us!

campbel:
Having trouble fixing the last check `DCO`

PR #35, opened 2020-08-25
Simplify State Management RFC

Pothulapati:
[Rendered RFC](https://github.com/linkerd/rfc/blob/tarun/simplify-state/design/0005-simplify-state-management.md)

## Summary

[summary]: #summary

Currently, Linkerd state in the `linkerd-config` ConfigMap as flags parsed from `installOptions`.
This has become counter intuitive after we added Helm support as `Values` struct
became our main source of truth (for rendering, etc), and causing back and forth
conversions between `installOptions` and `Values`. This RFC aims to remove the
unnecessary complexity in managing state and simplify things.

## Problem Statement (Step 1)

Currently, `linkerd-config` configMap is not only used for upgrades but also by
multiple components like Proxy Injector, CLI, etc to understand the current
cluster/linkerd state. This has the following problems:

- Lots of code debt/confusion to support the conversions between `installOptions`
and `Values`.
- As `Values` is not the source of truth, hacks [had to be/being](https://github.com/linkerd/linkerd2/pull/4864#pullrequestreview-467046753)
[made](https://github.com/linkerd/linkerd2/pull/4569) to make state (outside of flags, etc) to be stored and retrieved.
- No clear definition on which components use what configurations.

### DIFF
--- a/design/0005-simplify-state-management.md
+++ b/design/0005-simplify-state-management.md
@@ -13,27 +13,47 @@
 
 [summary]: #summary
 
-Currently, Linkerd state in the `linkerd-config` ConfigMap as flags parsed from `installOptions`. This has become counter intuitive after we added Helm support as `Values` struct became our main source of truth (for rendering, etc), and causing back and forth conversions between `installOptions` and `Values`. This PR aims to remove the unnecessary complexity in managing state and simplify things.
+Currently, Linkerd state in the `linkerd-config` ConfigMap as flags parsed from `installOptions`.
+This has become counter intuitive after we added Helm support as `Values` struct
+became our main source of truth (for rendering, etc), and causing back and forth
+conversions between `installOptions` and `Values`. This PR aims to remove the
+unnecessary complexity in managing state and simplify things.
 
 ## Problem Statement (Step 1)
 
-Currently, `linkerd-config` configMap is not only used for upgrades but also by multiple components like Proxy Injector, CLI, etc to understand the current cluster/linkerd state. This has the following problems:
+Currently, `linkerd-config` configMap is not only used for upgrades but also by
+multiple components like Proxy Injector, CLI, etc to understand the current
+cluster/linkerd state. This has the following problems:
 
-- As `Values` is not the source of truth, hacks [had to be/being](https://github.com/linkerd/linkerd2/pull/4864#pullrequestreview-467046753) [made](https://github.com/linkerd/linkerd2/pull/4569) to make state (outside of flags, etc) be stored for helm in some manner for Linkerd check.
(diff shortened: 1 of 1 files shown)


design/0005-simplify-state-management.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Design proposal (Step 2)
+
+[design-proposal]: #design-proposal
+
+This proposal aims to have a common way to store state both for the Linkerd2
+and Add-On Charts independently. This means that, each add-on/package gets its
+own configuration secret which is used for upgrades and to perform other CLI
+functions i.e Check, etc. Current CLI functions, etc will be updated to
+depend on this secret to understand the cluster state.
```
alpeb, line 43: Please add a note clarifying this new config secret will only be accessible from the CLI, and access won't be granted to any service account, except perhaps `linkerd-web` in order to perform `linkerd check` in the dashboard.

design/0005-simplify-state-management.md
```diff
@@ -0,0 +1,202 @@
...
+
+### For Linkerd
+
+For Linkerd state, There will be a new secret which contains overridden values
+which is a direct marshal of `Values` struct created from the `installOptions`
+passed through CLI flags. This `valuesOverride` will also contain any extra configuration
+created by the CLI i.e keys, etc other than the fields from `values.yaml`.
```
alpeb, line 50: Please note that this will be pure yaml, unlike the json strings we currently have in `linkerd-config`. Also it's probably not worth to back this with protobufs.

design/0005-simplify-state-management.md
```diff
@@ -0,0 +1,202 @@
...
+
+Currently, `linkerd-config` is used by most components to know cluster
+configuration(like `cluster-domain`, etc) and components like proxy-injector
+need a lot more. As the current `linkerd-config` will be removed,
+This information has to be passed to the components through flags.
+(flags are chosen over configMap, so that we don't fall into the trap of having
+more configuration resources/places). This makes each component get its configuration
+fully during the install/upgrade time through flags, without relying on external
+structures.
```
alpeb: Another nice side-effect is that this forces a component rollout when a config is changed through `linkerd upgrade`.

design/0005-simplify-state-management.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Summary
+
+[summary]: #summary
+
+Currently, Linkerd state in the `linkerd-config` ConfigMap as flags parsed from `installOptions`.
+This has become counter intuitive after we added Helm support as `Values` struct
+became our main source of truth (for rendering, etc), and causing back and forth
+conversions between `installOptions` and `Values`. This RFC aims to remove the
+unnecessary complexity in managing state and simplify things.
```
alpeb, line 20: I think it's worth framing the problem at first a little less technically, clarifying that the goal is to simplify state management for _upgrades_

olix0r:

PR #36, opened 2020-09-14
External Traffic Categorization: step one

n-oden:


n-oden:
So... this has been open w/o comment since 2020, and I think has been more or less entirely superseded by the new "mesh expansion" feature.  Long since time to put a fork in it, for better or worse.

PR #38, opened 2023-04-10
Rust profiling RFC

amit-62:
Profiling to Linkerd Rust controllers.


design/0005-rust-profiling.md
```diff
@@ -0,0 +1,32 @@
+# Title
+
+Contribution Name: (Profiling to Linkerd Rust controllers)
```
alpeb: # Add Profiling to the Linkerd Policy Controller

